PR #1, opened 2024-04-02
Add initial files.

brianorwhatever:
This adds the initial files for the typescript implementations of `did:tdw`


brianorwhatever:
@swcurran I can't add reviewers to this PR can I get a few more privileges on this repo 😃

swcurran:
Updated. Give it a go.

swcurran:
Happy to merge, but here are the things I found:

- No instructions for installing bun. Or running it from Docker if that is the right way to go :-).
- When I ran the test, it hung after completing the test, and I had to Ctrl-C to get back to the command line.  Weird...
- Two files were updated after running the test.
- When I ran the test with `:log` option the test completed (no hang) but now there were 13 files changed.

None of this end of world stuff — happy to approve if you think it best we go ahead and merge.

brianorwhatever:
- Yep good call will add instructions how to install bun as nobody will have it since it's so niche. Ideally everything can run in nodejs as well (albeit slower) but I'll have to test to double check that. I haven't created a dockerfile yet but will do that which will just be a resolver
- Ah yeah sorry that runs the watch version to rerun the tests after changes. That shouldn't be the default I'll swap in a regular run there and add a watch command
- Yeah I think at this point checking in the test output is good for comparisons and discussions but in the long run these should be ignored.

I'll tweak these things this evening and then we can merge in

brianorwhatever:
Reorganized the test commands and added a bit to the readme. I'll add a Dockerfile after I've worked on the resolver a bit more. I think we can merge this in now and follow up with a resolver PR

swcurran:
I'll leave you to merge this when you want -- feel free to update more or just merge.

brianorwhatever:
yep, `watch` is meant to hang and wait until files are changed and when that occurs run the tests again. `log` updates files as I've checked the output into git for now and everytime we run the tests it has a new timestamp so they will be updated. merging now.

swcurran:
Doh…got it.  Sorry — understand now.

PR #3, opened 2024-04-04
Added version and time queries

brianorwhatever:
Can now query versions by id or by time. Can also create and update at a specific time.. not sure if that's a generally good idea but was the easiest way to test the resolution features


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # trustdidweb-ts Changelog
 
+## 0.0.2 - 2024-04-04
+
+### Added
+- Add `options` to resolveDID.
+  - Option `versionIdQuery` to query specific version.
```
swcurran, line 7: Not sure if these should be standard DID Core query parameters?  versionId and versionTime — https://www.w3.org/TR/did-core/#did-parameters.  Or is there some indirection?

swcurran (approved):

brianorwhatever:
Updated to match DID core query parameters. Merging in now

PR #6, opened 2024-04-12
Use SCID as first log entry hash and use the proper cryptosuite

brianorwhatever:


swcurran (approved):

andrewwhitehead:
It seems like the digital bazaar library is using 'eddsa-2022' as the cryptosuite, I guess it's based on an earlier draft?

brianorwhatever:
yeah - I have removed that in this PR so now the only DB dependency is the multikey one used for key gen

PR #10, opened 2024-06-06
Adds prerotation capability

brianorwhatever:


swcurran (approved):

swcurran:
After running all the tests, the most of the logs don’t stay around, so I can’t see a test with pre-rotation.  How can I get an example that I can use in the spec?

Awesome stuff!

brianorwhatever:
Just made PR #11 that has a test that adds it to the log output from `bun run test:log`

PR #18, opened 2024-10-03
Witnesses

brianorwhatever:
This PR adds a witness implementation that works like this.
- When a DID is created/updated with a witness it sends a post request to a `/witness` endpoint that contains the DID log AFTER the controller has signed the update
- The witness endpoint will go through the verification methods (including private key) it has stored in an environment variable and if it finds one that is from a DID in the "witnesses" array of the log it has received it will sign a new proof
- The controller receives the witness proof in the response of the POST request


swcurran:
I see the CLI in the code — would love to have some instructions on how to use it.

So the witness in this case is built into the library and invoked directly via an endpoint, right?  Presumably in a “real world implementation”, we will probably have a separate repo (e.g. `trustdidweb-witness-ts`) and it would be tested with a bit of docker-compose magic to enable a separate implementation of the DID Controller functionality and the witness.  I’m not suggesting we do that right now — just asking if that is an appropriate longer term goal.

Any other feedback from the implementation?  We’ll meet again next Thursday — hope you can do a show and tell.

swcurran (approved):
As noted — would love to know how to do more than just run the tests — the CLI looks fun!

But verified that the tests run and things “look” good to me :-)

Fun stuff!

brianorwhatever:
Updated the readme a bit -- CLI is beta and experimental

swcurran (approved):

swcurran:
Awesome to have this CLI out there — can't wait to try it.  We need to get a video of using the CLI done — that would be excellent.

PR #22, opened 2024-11-07
Updates to the CLI

brianorwhatever:


swcurran (approved):
Looks good. I notice that the e2e CLI test fails because the witness server is not running. I assume that is expected?

```

Witness server is not running
Error fetching DID log: ConnectionRefused: Unable to connect. Is the computer able to access the url?
 path: "https://anywhy.ca/.well-known/did.jsonl"

ConnectionRefused: Unable to connect. Is the computer able to access the url?

```

I can confirm that I didn’t start a witness server. Is there a way to do that?

Great stuff!

brianorwhatever:
Yeah, so confusingly there is 2 different messages there. The "Witness server is not running" is actually fine - it will skip the witness tests in this case.
The error is actually coming from the interop test w/ @andrewwhitehead's DID. I'm not sure why that is happening as I can download that file right now.. I should actually remove the interop tests as this isn't the right place for them.. just was nice to have during development

brianorwhatever:
And to run a "witness server" it's the same as running a resolver `bun run server`

Issue #32, opened 2025-01-10
Resolve documents


brianorwhatever:
We should add document resolution generally before attempting to resolve did witness logs in #29

swcurran:
I think that resolving arbitrary, Controller added resources (e.g. AnonCreds objects, etc.) are different that the witness.json file, whose name and location are formally defined in the spec. The resources are for the most part defined outside the spec — in a document such as an “[AnonCreds Method](https://hyperledger.github.io/anoncreds-methods-registry/)”.  The witness.json file’s location and content are fully defined in the did:webvh spec.

That said, the did:webvh spec needs to define enough to enable defining a spec. like the “did:webvh AnonCreds Method.  I think we have enough now with the “DID URL Handling” section, but we’ll see in doing the AnonCreds Method spec.

brianorwhatever:
Ok, so what you're saying is we need to have a whitelist of documents from the spec? From a code perspective returning a `did-witness.json`, `did.jsonl` and for that matter `did.json` file run through relatively the same path. So in a server that is running didwebvh we want to only return documents as called out in the spec?

brianorwhatever:
hmm to me it seems arbitrary files can be resolved as shown in https://identity.foundation/didwebvh/#did-url-path-resolution-service

brianorwhatever:
Does `did:webvh:QmVt4QVuJbvMSuSHLae6kJwh9k9hZFV58yZZ4XKCxZa5RP:example.com/whois` fit into the `#files` pattern or "allow list"? The only difference with it is the `.vp` extension is currently dropped

swcurran:
[In the spec](https://identity.foundation/didwebvh/#whois-linkedvp-service), we have defined `/whois` as another implied service that uses the Linked-VP spec, so it is not part of the allow list. This gives the Controller flexibility on where they put it.

Reasonable?

brianorwhatever:
Closed in #35

PR #36, opened 2025-01-17
Witnesses updates

brianorwhatever:


swcurran (approved):
LGTM. I don’t see instructions on how to start the witness server, so I couldn’t run the witness tests.  Should that be possible?

FYI - also noticed that there is a file `bin/tdw` that I assume should be renamed?

brianorwhatever:
Yeah, apologies the witness server part of the tests was removed and I forgot to clean it up. Have removed it from the testing now as well.

yeah that `bin/tdw` will be updated in the next PR I create which will create the npm package. Closing #32 in favour of this PR and then merging this.

PR #50, opened 2025-03-11
Changes the crypto to use injected signer / verifier

brianorwhatever:


swcurran (approved):
LGTM.  Only oddity was running `bun run test` consitently gives errors retrieving the `anyway.ca` did.jsonl files, but I can access them using a browser, and using curl on the machine I’m running bun.

brianorwhatever:
@swcurran I had some very strange intermittent issues with that test as well.. not sure why but hoping the rest of the refactor might have removed those ghosts. Please review #51

Issue #53, opened 2025-03-28
Update historic version resolution results


brianorwhatever:
When resolving a past version, the entire log should be validated. If the log is valid, the most recent version metadata will be included in the resolution. If there is an invalid version in the chain, the last valid version details will be what is used.

PR #61, opened 2025-06-17
v1.0 updates

### LINKED ISSUE #57: Update implementation to v1.0
Other than changing the "method" in the `parameters`, I think the only change is to add Watchers.

The spec does not indicate who should notify a Watcher of updates -- the DID Controller or the Web Server serving the DID.  I assume that is an implementation detail, but it would be nice to know what the implementers think/recommend.

brianorwhatever:
Watcher support, internationalized domains, witness changes, bug fixes and resolution messages

Closes #57


brianorwhatever:
I'd like to fix #53 before this is merged in

swcurran:
> ohhh! that is a good point. I wasn't thinking about that case. On resolution it is just looking at the first log entry and assuming that is the version moving forward. Would you like me to implement this now or add an issue for making that update later?

If it not a huge stretch, it would be great to demonstrate that, and highlight it in the README and a test case.  I think it is an important feature for long lasting DIDs.

brianorwhatever:
hmmm thinking about this a bit more I think it would be a bit difficult. Reason being that currently the log is passed into the resolution process in it's entirety. To accommodate a multi version log we would need to rework how the processing is accomplished. Possible... but not trivial.

swcurran:
Sounds good. It is a pattern we want to support, but getting to v1.0 is more important right now.

swcurran:
Good stuff -- I helped!! :-)

I would say delete them before -- it's a lot of stuff if not cleared out.  If someone really wants them, they can grab them before re-running the test.

brianorwhatever:
done! merging now

Issue #66, opened 2025-06-30
Update the Resolution Document Metadata to make `threshold` a string


swcurran:
Per this [did:webvh spec PR](https://github.com/decentralized-identity/didwebvh/pull/229), the resolution document metadata in this implementation (and in the universal resolver) must be updated such that the `threshold` JSON value is the string equivalent of the `parameters -> witness -> threshold` integer.

This should be done as soon as is possible so that this and the Universal Driver implementation pass the DID Test Suite.

PR #67, opened 2025-06-30
Threshold uses string, update resolver witness file/proof handling, parameter handling

brianorwhatever:
As per #66 this makes the threshold return as a string. Also cleaned up some abstract class code.

### DIFF
--- a/src/assertions.ts
+++ b/src/assertions.ts
@@ -58,7 +58,7 @@ export const documentStateIsValid = async (
     proofs = [proofs];
   }
 
-  if (witness && witness.witnesses.length > 0) {
+  if (witness && witness.witnesses && witness.witnesses.length > 0) {
     if (!skipWitnessVerification) {
       validateWitnessParameter(witness);
     }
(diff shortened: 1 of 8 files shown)


swcurran:
I could be reading this wrong, but it looks to me like you changed the `threshold` in the `parameters` data to be a string (e.g. in the log entry), and that was not what we were going to do. Rather, we were just going to have the Resolution Document Metadata version of `threshold` to be string-ified. 

Am I misreading it?

swcurran:
Nothing in particular -- it just looked like there was code that was reading the data (shouldn't have to if just the Metadata) and the examples looked like it was the entry, not just the metadata.

Or I could be totally reading it wrong, and all is good...

brianorwhatever:
Ill give it another read through

swcurran:
Don't believe my comments -- I'm really just making sure that we don't change the contents of the Log.

swcurran (approved):
LGTM!

swcurran (approved):

swcurran (approved):

PR #74, opened 2025-08-28
fix: update DID without recalculating verification methods

genaris:
When developing a did:webvh registrar for credo-ts I've found that I couldn't make my DIDComm invitation services to work. Later I found that it was because, when updating the DID to add my custom services (including those for DID communication), this library was recalculating all verification method IDs. For multibase keys there was no problem, because it always generated the latest few chars of the public key. But for `Ed25519VerificationKey2018` and `X25519KeyAgreementKey2019` it was creating random values that made my services to lose the reference to them.

I guess this verification method normalization is necessary at the DID creation because we don't know the SCID beforehand, but when updating it we can allow passing verification methods as they are, we already know the full WebVH DID. But as I'm quite new to this spec, I'm not completely sure about that.

Comments and pointers to achieve my goal (add arbitrary defined verification methods in DID updates) are more than welcome!

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -402,30 +402,30 @@ export const normalizeVMs = (verificationMethod: VerificationMethod[] | undefine
   // First collect all VMs
   const vms = verificationMethod.map(vm => ({
     ...vm,
-    id: createVMID(vm, did)
+    id: vm.id ?? createVMID(vm, did)
   }));
   all.verificationMethod = vms;
 
   // Then handle relationships - default to authentication if no purpose is specified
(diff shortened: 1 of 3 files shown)


brianorwhatever:
I am away until Monday and will take a look closer then. Adding tests for the updating case might help show this functionality.

brianorwhatever (approved):
Looks good! Doing this in the normalize makes sense to me

genaris:
> @genaris [2.5.4](https://www.npmjs.com/package/didwebvh-ts/v/2.5.4) was released with this fix

Thanks!

PR #90, opened 2026-03-01
fix: resolver integrity and validation hardening

brianorwhatever:
- Fix hash chain to recompute entry hash instead of self-comparing versionId
  - Run hash chain validation on all entries including fast-resolve middle entries
  - Validate SCID from DID string matches SCID in fetched log
  - Fix empty nextKeyHashes array incorrectly enabling prerotation
  - Add publicKeyMultibase null check in verification method resolution
  - Add protocol version validation to v1.0 resolver
  - Surface error metadata on partial resolution failure
  - Fix CLI deactivate missing verifier
  - Remove legacy IGNORE_ASSERTION env var bypasses from assertion layer
  - Replace test scaffolding with real ed25519 verification throughout

### DIFF
--- a/src/method_versions/method.v1.0.ts
+++ b/src/method_versions/method.v1.0.ts
@@ -93,7 +93,7 @@ export const resolveDIDFromLog = async (log: DIDLog, options: ResolutionOptions 
   const resolutionLog = log.map(l => deepClone(l));
   const protocol = resolutionLog[0]?.parameters?.method;
   if (protocol !== PROTOCOL) {
-    throw new Error(`'${protocol}' protocol unknown.`);
+    throw new Error(`'${protocol}' is not a supported method version.`);
   }
   let did = '';
   let doc: any = null;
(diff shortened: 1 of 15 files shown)


src/method_versions/method.v1.0.ts
```diff
@@ -91,6 +91,10 @@ export const resolveDIDFromLog = async (log: DIDLog, options: ResolutionOptions
     throw new Error("Cannot specify both verificationMethod and version number/id");
   }
   const resolutionLog = log.map(l => deepClone(l));
+  const protocol = resolutionLog[0]?.parameters?.method;
+  if (protocol !== PROTOCOL) {
+    throw new Error(`'${protocol}' protocol unknown.`);
```
PatStLouis:
```suggestion
    throw new Error(`'${protocol}' is not a supported method version.`);
```

PatStLouis (approved):
LGTM, one small suggestion

brianorwhatever:
dang, using github ui to accept suggestion broke DCO. I'll fix that up tonight, merge this in and then update universal resolver

swcurran:
Yes -- to be safe, when I use the GitHub UI, I put the magic DCO string into the comment to make sure it works -- with both my emails since I'm not sure which one is being used...

PR #95, opened 2026-04-19
Fix prerotation key verification logic in method.v1.0.ts

swcurran:
Found a (possible?) bug when creating a compliance test suite via Claude. 

Fixes updateDID to validate the signer against params.updateKeys instead of lastMeta.updateKeys when pre-rotation is active, matching the existing resolver behaviour in resolveDIDFromLog.


swcurran:
@PatStLouis @brianorwhatever -- can you please give a review of this change?  If good, let's get it merged, a new release done and the universal resolver updated.

@brianorwhatever -- no need for you to do the steps, but would love it if you could tell us what to do for those two things (or point to the docs that tell us what to do).  Thanks!!

brianorwhatever:
Bug looks real. I have created #97 which just adds tests to this branch. Because I included [convential commit](https://www.conventionalcommits.org/en/v1.0.0/) as the commit title it will be picked up and released once it hits main which will trigger publishing a new version. universal resolver is as simple as bumping the version number [here](https://github.com/decentralized-identity/uni-resolver-driver-did-webvh)

brianorwhatever (approved):
once my PR is merged here looks good

PR #98, opened 2026-04-21
fix: default verification method controller to the DID

Dtitkaio:
Hi! First time contributing here. Thanks for maintaining this great library.

I hit this while using didwebvh-ts together with @digitalbazaar/vc, and noticed `normalizeVMs` doesn't default `controller` the way it already does for `id`. Happy to adjust the approach (e.g. throw instead of default) if you'd prefer.

## Summary

`normalizeVMs` currently spreads the input verification method as-is without defaulting `controller`, so DID documents produced by `createDID`/`updateDID` can be missing the `controller` property on verification methods.

Per [W3C DID Core §5.2](https://www.w3.org/TR/did-core/#verification-methods), `controller` is required on every verification method. `normalizeVMs` already defaults a missing `id` via `createVMID`; this PR extends the same pattern to `controller`.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,11 +1,11 @@
 import { canonicalize } from 'json-canonicalize';
 import { config } from './config';
-import { resolveDIDFromLog } from './method';
-import type { CreateDIDInterface, DIDDoc, DIDLog, VerificationMethod, WitnessProofFileEntry } from './interfaces';
 import { BASE_CONTEXT } from './constants';
-import { createBuffer, bufferToString } from './utils/buffer';
-import { createMultihash, encodeBase58Btc, MultihashAlgorithm } from './utils/multiformats';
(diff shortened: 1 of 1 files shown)


swcurran:
Thanks for the PR, but I'm confused on about it.  This seems to impact the DIDDoc. However, a did:webvh implementation should not be influencing the DIDDoc.  Is this in test code that generates a DIDDoc?

The only time (AFAIK) that the DIDDoc should be touched is:

- When submitting the initial DID, the DIDDoc should contain the placeholder ("{SCID}") wherever the DID exists in the DIDDoc. The initial DIDDoc will then have the calculated SCID where the placeholder exists.
- When resolving a DID, the "implicit services" are automagically added to the DIDDoc, if not already present.

How does the issue manifest?

brianorwhatever (approved):
Thanks!

PR #102, opened 2026-05-06
feat(dids): canonical did address encoding

rmlearney-digicatapult:
Enhanced DID address parsing for `createDID` with backwards `domain` compatibility in api and cli:

* Supports multiple input forms including `host`, `host:port`, `http://localhost` (local testing), `https://url`, `did:webvh:`
* Strict validation to reject malformed URLs, canonical DIDs and invalid ports, IP-address hosts
* Port-containing domains normalise to percent encoding (`%3A`) with rejection of double encoding

Modified test suite to cover wide range of acceptable and unacceptable address forms and resolver URL derivations

Same parsing behaviour used in both `v1.0` and `v0.5` paths

PR #103, opened 2026-05-07
feat(create): did document passthrough & did:web did document generation

rmlearney-digicatapult:
NB built on top of #102 

Add did document passthrough through union type in `createDID` to enable fully flexible enriched DID document creation for parity with Python and Rust implementations while preserving backwards compatibility.

Enables user to provide verification methods, services, key fragments, use `{DID}` and `{SCID}` as placeholders throughout for library to populate into final document.

Library able to automatically add alternative DID methods to the `alsoKnownAs` array through simple booleans
* `alsoKnownAsWeb: true` adds equivalent `did:web` URL and now generates a companion `did:web` did document defined by did:webvh v1.0 §3.7.10, returned in both `createDID()` and `updateDID()` as optionals for backwards compatibility during destructuring 

```ts
const { did, doc, meta, log, webDoc } = await createDID(...)
const { did, doc, meta, log, webDoc } = await updateDID(...)
```

Test suite for did document passthrough & did:web did document generation

**Noteworthy change** `alsoKnownAs` interface type previously required an array, this is now strictly enforced at runtime


PatStLouis (approved):

swcurran:
Ready to go, but needs to be resolved now that #102 has been merged.

Thanks!

swcurran:
Wondering if we need to add documentation about this capability?

PR #104, opened 2026-05-08
feat: accept ecdsa-jcs-2019 + P-256 multikey in proof verification

juliaxgulia:
## Summary

The W3C VC Data Integrity spec defines multiple ECDSA cryptosuites — `eddsa-jcs-2022` for Ed25519 and `ecdsa-jcs-2019` for P-256 — over the same JCS canonicalization. Today this library accepts only `eddsa-jcs-2022`, so `did:webvh` logs minted with P-256 keys are unresolvable through `resolveDIDFromLog` and `verifyWitnessProofs`, even though the underlying hash + signature primitives are interchangeable and the `Verifier` hook is already algorithm-agnostic.

This PR extends three checks in the verify path to accept the P-256 path alongside Ed25519. **Backward-compatible** — the new conditions are strictly broader; every existing Ed25519 log continues to verify unchanged. **No API change** — the `Verifier.verify(signature, message, publicKey)` hook stays the same; consumers pick the right ECDSA/EdDSA primitive in their own implementation.

## What changed

`src/assertions.ts` (`documentStateIsValid`):

- Cryptosuite check now accepts `eddsa-jcs-2022` **or** `ecdsa-jcs-2019`.
- Multikey-header byte check now accepts Ed25519 `[0xed, 0x01]` **or** P-256 `[0x80, 0x24]` (varint-encoded multicodec `0x1200` for `p256-pub`).

`src/witness.ts` (`calculateWitnessWeight` + `verifyWitnessProofs`):

- Two cryptosuite checks extended the same way.
- Public-key length check now accepts `34` (Ed25519: 2-byte prefix + 32-byte raw) **or** `35` (P-256: 2-byte prefix + 33-byte SEC1-compressed).

## Why this is safe

- The `Verifier` interface (`src/interfaces.ts`) is already algorithm-agnostic: `verify(signature, message, publicKey)`. Accepting a P-256 cryptosuite doesn't require any verifier change in the library — the consumer's `verify` implementation does the right thing for the public key it sees.
- The signing side (`createProof`, `method.v0.5`, `method.v1.0`, `cli.ts`) is **not** touched in this PR. Those still hardcode `cryptosuite: 'eddsa-jcs-2022'` for compatibility, and signers that want to mint P-256 logs continue to do so via their own `Signer` implementation (as before, just now with the reciprocal verify path supported by this library).
- The existing test suite for Ed25519 keeps passing; no test fixture is changed.

## How it was tested

The change has been running in production at Raisonnai (an art catalogue + identity-infrastructure project that uses `did:webvh` for P-256-signed verifiable credentials). Our integration test now exercises a full `prepareGenesisLogEntry → finalizeGenesisLogEntry → resolveDIDFromLog` round-trip with P-256 keys against this patch — it had previously been skipped with a `// TODO: library hardcodes eddsa-jcs-2022` comment.

Happy to add an in-tree P-256 test if you'd like — that would mean a new dev-dep (e.g. `@noble/curves`), so I left it out to keep this PR minimal. Let me know your preference.

## Notes

- A future PR could extend `createProof` and the v1.0/v0.5 method implementations to mint P-256 logs from this library directly (currently consumers compose the signing themselves). That's larger and orthogonal — happy to follow up if it's wanted, but separating the two felt right since this PR's goal is to unblock cross-implementation interop on the *verify* side.

### DIFF
--- a/src/assertions.ts
+++ b/src/assertions.ts
@@ -76,7 +76,7 @@ export const documentStateIsValid = async (
     if (proof.proofPurpose !== 'authentication' && proof.proofPurpose !== 'assertionMethod') {
       throw new Error(`Unknown proof purpose ${proof.proofPurpose}`);
     }
-    if (proof.cryptosuite !== 'eddsa-jcs-2022') {
+    if (proof.cryptosuite !== 'eddsa-jcs-2022' && proof.cryptosuite !== 'ecdsa-jcs-2019') {
       throw new Error(`Unknown cryptosuite ${proof.cryptosuite}`);
     }
 
@@ -86,8 +86,15 @@ export const documentStateIsValid = async (
     }
 
     const publicKey = multibaseDecode(vm.publicKeyMultibase).bytes;
-    if (publicKey[0] !== 0xed || publicKey[1] !== 0x01) {
-      throw new Error(`multiKey doesn't include ed25519 header (0xed01)`);
+    // Multikey multicodec prefixes (varint-encoded):
+    //   Ed25519 (0xed)  → [0xed, 0x01]
+    //   P-256   (0x1200) → [0x80, 0x24]
+    // The Verifier hook is algorithm-agnostic — the consumer's `verify`
+    // implementation is responsible for picking the right primitive.
+    const isEd25519MultiKey = publicKey[0] === 0xed && publicKey[1] === 0x01;
+    const isP256MultiKey = publicKey[0] === 0x80 && publicKey[1] === 0x24;
+    if (!isEd25519MultiKey && !isP256MultiKey) {
+      throw new Error(`multiKey doesn't include ed25519 (0xed01) or p256 (0x8024) header`);
     }
 
     const {proofValue, ...restProof} = proof;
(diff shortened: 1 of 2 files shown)


swcurran:
The addition of other cryptosuites to authorize or witness a did:webvh DID update is not permitted in the did:webvh spec -- only the eddsa-jcs-2022 cryptosuite is permitted. The spec defining exactly what cryptography is permitted is done on purpose to make interop easier. Verifier of did:webvh Log Entries know from the spec exactly what they have to support.

Note that DIDDoc added to any did:webvh DID Log can use any cryptographic schemes as decided by the DID Controller.  So a DID Controller can add a `verificationMethod` to their DIDDoc with a P-256 key. The limitation is only on the keys used to sign the proofs to authorize the update (the `proof` field in the DID Log entry) and the witness proofs.

We expect that a PQC cryptosuite, likely using ML-DSA 44, will be added to the did:webvh spec in the next major version. This will enable support for either EDDSA or ML-DSA 44 keys to be used for authorizing or witnessing a did:webvh DID Log.

Thanks for the PR, but closing this as being non-conformant.  Hope that makes sense.

PR #110, opened 2026-05-20
fix: publish via npm OIDC trusted publishing instead of static token

brianorwhatever:
The pre-#107 semantic-release-based workflow authenticated to npm via OIDC trusted publishing ("Verifying OIDC context for publishing from GitHub Actions" / "OIDC token exchange with the npm registry succeeded"). The static NPM_TOKEN_WEBVH_TS path adopted after #107 fails with EOTP because the token type does not bypass 2FA on this package.

Restore the previous auth pattern by dropping NODE_AUTH_TOKEN and running publish through npx with a recent npm (the npm shipped with node 22 is too old for native OIDC trusted publishing; an earlier attempt to upgrade the global npm in-place hit the known self-upgrade MODULE_NOT_FOUND bug, so npx is used instead). The id-token: write permission was already in place for provenance.

The Trusted Publisher configuration on npmjs.com must continue to match this repo and .github/workflows/publish.yml.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -58,6 +58,4 @@ jobs:
         run: bun run build
 
       - name: Publish to npm
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN_WEBVH_TS }}
-        run: npm publish --provenance
+        run: npx -y npm@latest publish --provenance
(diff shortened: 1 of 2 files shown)


swcurran (approved):

swcurran:
@brianorwhatever -- conflicts in this PR.

PR #112, opened 2026-05-22
feat(core): proof template

rmlearney-digicatapult:
Improve type safety and spec alignment

* Introduce new interfaces for `DataIntegrityProofTemplate` and `DataIntegrityProofPurpose` and use these for API boundary operations and in the CLI when building compliant Data Integrity proof objects

* Tightened `document` in `SigningInput` from `any` to `unknown`

* Witness verification now validates `proofPurpose === assertionMethod` in accordance with spec

* DID log proof purpose is now strictly `assertionMethod` with missing-proof guard

**Breaking 1**

Consumers using **deep imports** of `createWitnessProof`

`createWitnessProof` (internal use) now requires explicit `verificationMethod` to match Data Integrity spec requirements

Old:
`(doc: any, proofTemplate?: any)`

New:
`(doc: { versionId: string }, proofTemplate?: DataIntegrityProofTemplate)` where `DataIntegrityProofTemplate` requires `verificationMethod`

**Breaking 2**

API users passing looser proof objects into signing functions

* Stricter types in `interfaces.ts` could cause breaks for some consumers passing proof shapes into `signer` without `verificationMethod`

* `type` and `cryptosuite` are now exact literals not string (`DataIntegrityProof` and `eddsa-jcs-2022` respectively)

### DIFF
--- a/src/witness.ts
+++ b/src/witness.ts
@@ -1,16 +1,19 @@
 import { canonicalize } from 'json-canonicalize';
 import { createHash } from './utils/crypto';
-import type { DataIntegrityProof, DIDLogEntry, WitnessEntry, WitnessProofFileEntry, Verifier, WitnessParameterResolution } from './interfaces';
+import type { DataIntegrityProof, DataIntegrityProofTemplate, DIDLogEntry, WitnessEntry, WitnessProofFileEntry, Verifier, WitnessParameterResolution } from './interfaces';
 import { resolveVM } from "./utils";
 import { concatBuffers } from './utils/buffer';
 import { fetchWitnessProofs } from './utils';
 import { multibaseDecode } from './utils/multiformats';
 
 export async function createWitnessProof(
-  signer: (doc: any, proofTemplate?: any) => Promise<{proof: any}>,
+  signer: (
+    doc: { versionId: string },
+    proofTemplate?: DataIntegrityProofTemplate
+  ) => Promise<{ proof: Partial<DataIntegrityProof> }>,
   versionId: string
 ): Promise<DataIntegrityProof> {
-  const proofTemplate = {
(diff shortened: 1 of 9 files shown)


swcurran:
Ran the didwebvh test suite on the PR and no regressions identified. Good to go from that perspective.

src/witness.ts
```diff
@@ -79,6 +96,14 @@ export async function verifyWitnessProofs(
   for (const proofSet of witnessProofs) {
     // Process each proof in the set
     for (const proof of proofSet.proof) {
+      if (proof.type !== 'DataIntegrityProof') {
+        throw new Error('Invalid witness proof type');
+      }
+
+      if (proof.proofPurpose !== 'assertionMethod') {
```
brianorwhatever, line 103: This will cause this library to invalidate witness proofs created in previous versions of this library and I'm not sure the spec mentions the proofPurpose for witness proofs. The spec should probably be explicit here @swcurran

brianorwhatever:
This looks good and the tightening of types is a great change. I'm just not sure on the `authentication` -> `assertionMethod` change for witness proofs introduced

src/witness.ts
brianorwhatever: nevermind found it..

> proof is an array of [Data Integrity](https://identity.foundation/didwebvh/v1.0/#term:data-integrity) proofs that use the versionId as input data. The permitted [Data Integrity](https://identity.foundation/didwebvh/v1.0/#term:data-integrity) cryptosuites used by the [witnesses](https://identity.foundation/didwebvh/v1.0/#term:witnesses) MUST be eddsa-jcs-2022 as referenced in [spec:di-eddsa-v1.0] and the proofPurpose set to assertionMethod.

ok so we do need this breaking change

brianorwhatever (approved):

PR #113, opened 2026-05-22
feat(witness): witness API endpoints and did:key handling

rmlearney-digicatapult:
Implemented API endpoints (wired to CLI) for witness interactions & hardened did:key handling

### New API methods:

* `createWitnessProof` (create and sign one witness proof for a specific `versionId`)
* `signWitnessProofEntry` (sign one witness proof entry for a specific `versionId`)
* `signWitnessProofEntries` (batch signing of multiple entries)

### Witness threshold

Implemented witness thresholding, accepting failing witnesses per spec

### did:key

`did:key` handling now has strict parser helpers for DID-form versus verification method form in `utils.ts` to match spec requirements. Inspired by did:key handling in rust and python codebases.

* `parseDidKeyDid`
* `parseDidKeyVerificationMethod`

Witness identity now uses exact parsed DID equality (not prefix matching), each witness DID parsed strictly when presented

### did:webvh

Removed `did:webvh` as shortcut witness authorisation path for DID log entry proofs to align with strict spec requirement for key-based (`did:key`) verificationMethod

### Error types

Error types now have strict enum for supported error types with spec-aligned reporting in witness paths

### Other

`fastResolve` is now a user-accessible option through the API, defaulting to `false` for spec compliance

### Test coverage

Tests for e2e witness lifecycle coverage

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -31,6 +31,11 @@ export interface Signer {
   getVerificationMethodId(): string;
 }
 
+export interface WitnessSigner {
+  sign(input: SigningInput): Promise<SigningOutput>;
+  getVerificationMethodId(): string;
+}
+
 export interface Verifier {
   verify(signature: Uint8Array, message: Uint8Array, publicKey: Uint8Array): Promise<boolean>;
 }
@@ -91,6 +96,24 @@ export interface VerificationMethod {
 
 export interface WitnessEntry {
   id: string;  // did:key DID
+}
+
+export interface ParsedDidKeyVerificationMethod {
+  did: string;
+  fragment?: string;
+  keyMultibase: string;
+}
+
+export interface WitnessSigningOptions {
+  versionId: string;
+  witnesses: WitnessEntry[];
+  witnessSignersByDid: Record<string, WitnessSigner>;
+  created?: string;
+}
+
+export interface WitnessSigningResult {
+  versionId: string;
+  proof: DataIntegrityProof[];
 }
 
 export interface WitnessParameter {
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -31,11 +31,6 @@ export interface Signer {
   getVerificationMethodId(): string;
 }
 
-export interface WitnessSigner {
-  sign(input: SigningInput): Promise<SigningOutput>;
-  getVerificationMethodId(): string;
-}
-
 export interface Verifier {
   verify(signature: Uint8Array, message: Uint8Array, publicKey: Uint8Array): Promise<boolean>;
 }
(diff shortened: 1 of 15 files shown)


swcurran:
Ran the branch through the test suite and all good -- with one negative test ("negative-cross-did-witness-replay") changed from FAIL to PASS (yay!).

I think the fast-resolve option needs to be removed unless it works via having the existing verification state available and the "fast" processing is only checking the added log entries -- e.g. there is a passed in cache of the existing state.  But that is for another PR.  I like the defaulting to false for fast resolve.

@PatStLouis @brianorwhatever -- over to you for final review.

src/interfaces.ts
```diff
@@ -12,6 +31,11 @@ export interface Signer {
   getVerificationMethodId(): string;
 }
 
+export interface WitnessSigner {
```
brianorwhatever: this is identical to `Signer` we should probably just reuse that?

brianorwhatever:
`countVerifiedWitnessApprovals` wraps each proof in `try { … } catch { continue }`. This silently swallows invalid-cryptosuite, unauthorized-witness, and malformed proofs that the old `verifyWitnessProofs` *hard-rejected*. The DID still gets rejected when the threshold isn't otherwise met, but the failure mode is now indistinguishable from "no proofs," and an attacker mixing malformed proofs with enough valid ones gets the bad ones silently dropped. I think we should at minimum log the swallowed error. Note this also leaves `verifyWitnessProofs` as dead code in v1.0 (still live in v0.5).

brianorwhatever:
Let's fix these merge conflicts up, maybe add a log in the bad witness catch and handle the dead code and this should be good to go

src/interfaces.ts
rmlearney-digicatapult: Have reused `Signer`, have a follow-up branch which narrows for type safety

rmlearney-digicatapult:
> `countVerifiedWitnessApprovals` wraps each proof in `try { … } catch { continue }`. This silently swallows invalid-cryptosuite, unauthorized-witness, and malformed proofs that the old `verifyWitnessProofs` _hard-rejected_. The DID still gets rejected when the threshold isn't otherwise met, but the failure mode is now indistinguishable from "no proofs," and an attacker mixing malformed proofs with enough valid ones gets the bad ones silently dropped. I think we should at minimum log the swallowed error. Note this also leaves `verifyWitnessProofs` as dead code in v1.0 (still live in v0.5).

Error now logged, updated v0.5 to use new counting logic and removed dead `verifyWitnessProofs`

PatStLouis (approved):

PR #114, opened 2026-05-22
fix(types): add type checking script and fix errors in tests

rmlearney-digicatapult:
First pass to enforce type safety

Implemented `bunx tsc --noEmit` as package script

Fixed errors in tests with strict non-null assertions, enum comparison

Expanded `tsconfig` to include scripts

NB merge after #113

### DIFF
--- a/src/method_versions/method.v1.0.ts
+++ b/src/method_versions/method.v1.0.ts
@@ -1,6 +1,7 @@
 import { createDate, createDIDDoc, createSCID, deriveHash, findVerificationMethod, getActiveDIDs, getBaseUrl, replaceValueInObject, deepClone, enrichAlsoKnownAs, generateParallelDidWeb, parseCanonicalAddress, replaceCreateDidPlaceholders, validateCreateDidDocument } from "../utils";
 import { METHOD, PLACEHOLDER } from '../constants';
 import { documentStateIsValid, hashChainValid, newKeysAreInNextKeys, scidIsFromHash } from '../assertions';
+import { DidResolutionError } from '../interfaces';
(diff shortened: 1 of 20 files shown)


src/method_versions/method.v1.0.ts
```diff
@@ -533,4 +561,60 @@ export const deactivateDID = async (options: DeactivateDIDInterface & { updateKe
...
+  targetVersionNumber: number;
+  witness: WitnessParameterResolution;
+}
+
+const getEntryWitnessParameter = (parameters: DIDLogEntry['parameters']): WitnessParameterResolution | undefined => {
+  if ('witness' in parameters) {
+    return parameters.witness ?? {};
+  }
+
+  if ((parameters as any).witnesses) {
```
brianorwhatever, line 588: these `(parameters as any)` should probably be typed with a proper interface as well

brianorwhatever:
This PR is great but appears to recreate a bunch of #113 which do we want to merge?

src/method_versions/method.v1.0.ts
rmlearney-digicatapult: I've had to purposely ignore `any` in this first pass because there are just too many of them. I've got them cleaned up in a local branch which I'll push after sequential merges of these PRs (they all build incrementally)

rmlearney-digicatapult:
> This PR is great but appears to recreate a bunch of #113 which do we want to merge?

Apologies, I've been building by locally forking latest developments rather than basing directly on main which means each one drags along history. The plan was to put these up to show what I'm working on, then sequentially rebase as the older PRs are merged in

PatStLouis:
@rmlearney-digicatapult looks like this pr has a merge conflict.

rmlearney-digicatapult:
@PatStLouis fixed :)

brianorwhatever (approved):

PR #115, opened 2026-05-28
chore(linter): implement Biome linter, first fixes and static tests in CI

rmlearney-digicatapult:
Installed `Biome` with custom settings to match existing repo majority patterns:
* space indents
* trailing commas
* semicolon line terminations
* single quotations
* ignore unused variables and function parameters
 
Different strictness levels for tests/scripts/examples versus core library

Turned off `any` checking in first pass but implemented null assertion fixes, chain operator, and cleaned up one implicit `any` and one test file use of `any`

Introduced `ci.yml` to run static checks (types, linting, in-repo tests) in CI

Lots of files touched in this first linting fix


brianorwhatever:
This PR is great but includes the functionality changes from the other PRs. Let's get them in first and then we can do this one

rmlearney-digicatapult:
@brianorwhatever yes apologies I've been building sequentially on top of previous PRs and pushing for visibility, with plan to rebase on each merge to clean up

Will leave as `draft` until ready to go

PatStLouis:
@rmlearney-digicatapult theres a couple failing tests and merge conflict

rmlearney-digicatapult:
@PatStLouis fixed :)

PatStLouis (commented):
## Follow-up: format CI step + pinning

Two non-blocking items worth fixing before or right after merge.

### 1. Misleading “Format check” step

CI currently runs:

```yaml
- name: Lint
  run: bun run lint          # → biome check .

- name: Format check
  run: bun run format        # → biome format .
```

These are not equivalent:

| Command | What it does | Fails CI if code is unformatted? |
|---------|----------------|-----------------------------------|
| `biome check .` | Lint **and** verify formatting | **Yes** |
| `biome format .` | Print what formatted code would look like | **No** |

So the step named **Format check** does not actually enforce formatting. Enforcement already happens in **Lint** via `biome check`.

**Suggested fix (pick one):**
- Remove the redundant format step and rely on `biome check`, or
- Change it to `bunx biome format --check .` (or consolidate on `biome ci .`).

Not a functional bug today — lint already catches format drift — but the workflow label doesn’t match behavior.

### 2. Pin Bun and Biome schema for CI stability

**Bun:** `bun-version: latest` in `.github/workflows/ci.yml` means CI can change behavior when Bun ships a new release, even on the same commit. Pin to a known version (e.g. match local dev) and bump intentionally.

**Biome schema:** `biome.json` uses `"$schema": ".../schemas/latest/schema.json"`. That can drift from the installed `@biomejs/biome` version in the lockfile and cause editor/CLI mismatches. Prefer pinning to the lockfile version, e.g. `.../schemas/2.4.16/schema.json`.

Neither blocks the core goal of enforcing lint/format/types/tests in CI — overall this PR looks good once those are addressed.

PatStLouis (commented):
## Correction + shorter follow-up

**Correction:** `biome format .` (without `--write`) *does* fail CI on unformatted files — I was wrong to call that step non-enforcing. There is no `--check` flag; omitting `--write` is the check mode.

**What still applies:**

1. **Redundant CI step** — `bun run lint` (`biome check .`) already verifies formatting *and* lint. The separate “Format check” step only re-runs format. Suggest dropping it or replacing both with `biome ci .`.

2. **Pin versions** — `bun-version: latest` and `biome.json` `"$schema": ".../latest/..."` can drift from the lockfile and cause surprise CI/editor mismatches. Pin both to the versions in `bun.lock`.

Non-blocking; otherwise looks good.

PatStLouis (commented):
## CI follow-up (non-blocking)

1. **Redundant format step** — `bun run lint` (`biome check .`) already checks formatting and lint. The separate “Format check” step (`biome format .`) only re-runs format. Suggest dropping it or replacing both with `biome ci .`.

2. **Pin versions** — `bun-version: latest` and `"$schema": ".../latest/..."` in `biome.json` can drift from the lockfile. Pin both to the versions in `bun.lock`.

Otherwise looks good.

swcurran:
As I think you were expecting, more conflicts after the merging of the canonicalization PR. Back to you @rmlearney-digicatapult

rmlearney-digicatapult:
@PatStLouis fixed :)

swcurran:
All good with the test suite!

brianorwhatever (approved):
Lots of changes but I think it’s all linting 👍

PR #116, opened 2026-05-28
fix(json): apply strict canonicalisation

rmlearney-digicatapult:
Bumped dependencies and implemented strict runtime JSON canonicalisation to enforce RFC 8785 adherence and behavioural parity with py and rs implementations

* error on infinite values, circular references, undefined in arrays
* strip `undefined` fields for javascript normalisation (not part of RFC) but preserve `null` with semantic distinction from absent values

Witness signer output with undefined object fields is accepted and sanitised

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,104 +5,52 @@
     "": {
       "name": "didwebvh-ts",
       "dependencies": {
-        "@noble/hashes": "^1.8.0",
+        "@noble/hashes": "^2.2.0",
         "cookie": "^1.1.1",
-        "glob": "^13.0.0",
+        "glob": "^13.0.6",
         "js-yaml": "^4.1.1",
-        "json-canonicalize": "^1.0.6",
+        "json-canonicalize": "^2.0.0",
       },
       "devDependencies": {
-        "@semantic-release/changelog": "^6.0.3",
-        "@semantic-release/git": "^10.0.1",
-        "@sinclair/typebox": "0.32.35",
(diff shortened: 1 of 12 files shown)


rmlearney-digicatapult:
I think I'd like to swap `json-canonicalize` for `canonicalize` in a future PR which seems to have stricter RFC 8785 compliance and is better maintained

PatStLouis:
@rmlearney-digicatapult couple merge conflicts to address

swcurran:
I tried to run the branch in the in the https://github.com/decentralized-identity/didwebvh-test-suite 's new Docker model and I got an error. Asked Claude about it and it says the error is in this PR. Here's the feedback:

> The fix/canonicalization branch on rmlearney-digicatapult/didwebvh-ts has a type error in witness.ts — the object literal is missing a required type field on DataIntegrityProof. That causes bun run build to exit with code 1, which fails the Docker layer.
>
> Nothing is wrong with our script or Dockerfile. You'd need to let the PR author know their branch doesn't build cleanly — they need to fix the TypeScript error in witness.ts:86 before the branch can be tested here.

Please take a look!

swcurran:
Nice -- that fixed it.  And nice that I was able to run the test suite again to see it worked :-). If I see more commits, I'll keep running it...

I'd still prefer a dev add an approval, so hoping that @PatStLouis or @brianorwhatever could do a check on this one.

Thanks @rmlearney-digicatapult !

PatStLouis (approved):
LGTM

PR #119, opened 2026-06-09
fix(core): type safety

rmlearney-digicatapult:
The big one to clean up all uses of the `any` type and ensure type safety throughout before moving on to further spec alignment work

**Type Safety**

* enforce stricter linting to `error` on use of `any` type
* tighten signer/input/output and DID document typing
* Improve unknown-safe error handling and utility contracts
* Update tests with stricter typing and fewer permissive casts

**Other**

* changed error timing/messages in CLI and resolver fail-fast behaviour
* DID return paths now enforce presence of valid DID id
* v1.0 resolution now fails fast when no valid DID document can be resolved
* CLI fails earlier when required key material is missing
* CLI `update --updateKey` now throws when no matching verification method can be found
* FS loading uses `node:fs` as primary with compatibility fallback for runtimes/bundlers which reject `node:fs`

**witness.ts**

* return object for `createWitnessProof()` uses optional fields from `sanitizedProof` and required validated string fields from `mergedProof`

**legacyParameters**

This is for backwards resolution compatibility with existing log files and is cast during resolution in the `method.v*.ts` files

```ts
legacyParameters: {
  witnesses?: { id: string }[]
  witnessThreshold?: string | number
}
```

```ts
parameters: {
  method?: string
  scid?: string
  updateKeys?: string[]
  nextKeyHashes?: string[]
  portable?: boolean
  witness?: {
    threshold?: number
    witnesses?: { id: string }[]
  }
  watchers?: string[] | null
  deactivated?: boolean
}
```

We could introduce a hard-breaking change + a migration script for old did logs

### DIFF
--- a/src/method_versions/method.v1.0.ts
+++ b/src/method_versions/method.v1.0.ts
@@ -5,10 +5,12 @@ import type {
   CreateDIDResult,
   DataIntegrityProof,
   DeactivateDIDInterface,
+  DIDDoc,
   DIDLog,
   DIDLogEntry,
   DIDResolutionMeta,
   ResolutionOptions,
+  ServiceEndpoint,
   UpdateDIDInterface,
   UpdateDIDResult,
   WitnessParameterResolution,
@@ -76,7 +78,7 @@ export const createDID = async (options: CreateDIDInterface): Promise<CreateDIDR
     return vm;
   });
 
-  let doc: any;
+  let doc: DIDDoc;
   if (options.didDocument) {
     validateCreateDidDocument(options.didDocument);
     doc = deepClone(options.didDocument);
@@ -169,7 +171,7 @@ export const createDID = async (options: CreateDIDInterface): Promise<CreateDIDR
 export const resolveDIDFromLog = async (
   log: DIDLog,
   options: ResolutionOptions & { witnessProofs?: WitnessProofFileEntry[]; fastResolve?: boolean } = {}
-): Promise<{ did: string; doc: any; meta: DIDResolutionMeta }> => {
+): Promise<{ did: string; doc: DIDDoc; meta: DIDResolutionMeta }> => {
   if (options.verificationMethod && (options.versionNumber || options.versionId)) {
     throw new Error('Cannot specify both verificationMethod and version number/id');
   }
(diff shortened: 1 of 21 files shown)


swcurran:
Ran this through the test suite and all good so I'm good with it.

@brianorwhatever and/or @PatStLouis please take a look.

src/method_versions/method.v1.0.ts
```diff
@@ -225,13 +227,13 @@ export const resolveDIDFromLog = async (
       if (version === '1') {
         meta.created = versionTime;
         newDoc = state;
-        host = newDoc.id.split(':').at(-1);
-        meta.scid = parameters.scid;
+        host = requireDidId(newDoc.id).split(':').at(-1) ?? '';
```
brianorwhatever, line 230: do we have any guards that ensure the last item (-1) is the host? Worried there might be cases where a DID with an extra colon makes it through to here and messes with things. IE `did:webvh:Qz4a…:example.com:extra`

It’s clearly malformed so hopefully that is caught before here

brianorwhatever (approved):

src/method_versions/method.v1.0.ts
swcurran: @brianorwhatever -- that DID is valid.  It means that the did.jsonl resides at example.com/extra/did.jsonl.

swcurran:
I'm going to go with the approval and merging this one.  Hope that works @brianorwhatever .

src/method_versions/method.v1.0.ts
brianorwhatever: d'oh yes of course.. been awhile. so this is a bigger concern then. it won't grab the host it will grab the last path of a DID like that OR the host of a bare domain did

swcurran:
Dang...and I've merged it. Please @rmlearney-digicatapult or @brianorwhatever -- please provide a correction.  Sorry to be a pain...

brianorwhatever:
all good yeah I'll take a look

rmlearney-digicatapult:
Will do

rmlearney-digicatapult:
Sorry I forgot, on the next PR (`fix/negative-tests`) this is now handled by the new `parseCanonicalAddress()` which parses the DID properly and splits by parts so that extra colon segments become path components

PR #120, opened 2026-06-10
fix(core): add guards to address failing negative tests

rmlearney-digicatapult:
Following @swcurran's report of failing negative test cases in `didwebvh-test-suite` and reviewed against the `webvh` spec.

Normative fixes:

* Added ISO8601 utility & enforce `versionTime` validity and monotonic progression in logs
* Support only `Z` or `+00:00` for versionTime
* Permit 5 minute skew
* Enforced `SCID` immutability in portability updates and absence from non-initial logs
* Implemented structured DID identifier parsing for `did:webvh`
* Normalise/decode host before IP checks
* Do not ignore encoding case for IP rejection
* Removed `fastResolve` functionality
* Removed `localhost` special-casing for HTTPS

Security hardening

* Reject dot-segments and encoded traversal semantics in DID path segments
* Reject decoded slash inside a single DID path segment

### DIFF
--- a/src/method_versions/method.v1.0.ts
+++ b/src/method_versions/method.v1.0.ts
@@ -32,6 +32,11 @@ import {
   replaceValueInObject,
   validateCreateDidDocument,
 } from '../utils';
+import {
+  createNextVersionTime,
+  parseUtcIso8601VersionTime,
+  validateUtcIso8601NotInFuture,
+} from '../utils/iso8601-datetime';
 import { countVerifiedWitnessApprovals, fetchWitnessProofs, validateWitnessParameter } from '../witness';
 
 const VERSION = '1.0';
@@ -64,6 +69,9 @@ export const createDID = async (options: CreateDIDInterface): Promise<CreateDIDR
   const allPaths = [...(parsed.paths || []), ...(options.paths || [])];
   const path = allPaths.length > 0 ? allPaths.join(':') : undefined;
   const controller = `did:${METHOD}:${PLACEHOLDER}:${didDomainComponent}${path ? `:${path}` : ''}`;
+  if (options.created) {
+    validateUtcIso8601NotInFuture(options.created, 'createDID created');
+  }
   const createdDate = createDate(options.created);
 
   // Safety guard: Strip secret keys from verification methods before creating DID document
@@ -201,6 +209,8 @@ export const resolveDIDFromLog = async (
   let lastValidMeta: DIDResolutionMeta | null = null;
   let i = 0;
   let host = '';
+  let previousVersionTime: Date | undefined;
+  const resolutionNow = new Date();
   const requiredWitnessChecks: RequiredWitnessCheck[] = [];
 
   // Fast resolution is opt-in; full verification is the default conformant path.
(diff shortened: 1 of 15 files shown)


swcurran:
Confirmed the fixes complete the test suite successfully and address the negative tests in the suite.

src/utils/iso8601-datetime.ts
```diff
@@ -0,0 +1,122 @@
...
+ * Enforces:
+ * - Strict ISO 8601 format with calendar correctness (via regex)
+ * - Timezone component required (Z or ±HH:MM offset)
+ * - Semantic date validity (via Date.parse)
+ *
+ * Spec reference: did:webvh v1.0 §3.5 and §3.6.2 (`versionTime` in UTC ISO8601)
+ * https://identity.foundation/didwebvh/v1.0/
+ */
+export function parseUtcIso8601VersionTime(value: string, context: string): Date {
+  const match = ISO8601_DATETIME_REGEX.exec(value);
```
brianorwhatever, line 84: The regex permits any `±HH:MM` offset, so this accepts non-UTC timestamps like `2025-01-01T00:00:00+05:00`.

> The versionTime MUST be a valid UTC [ISO8601](https://identity.foundation/didwebvh/v1.0/#term:iso8601) string with explicit Z (or +00:00); values without a zone designator, or expressing a non-UTC zone, MUST be rejected.

The spec is explicit the offset must be Z

brianorwhatever: In https://identity.foundation/didwebvh/v1.0/#read-resolve

src/method_versions/method.v1.0.ts
```diff
@@ -402,6 +417,10 @@ export const resolveDIDFromLog = async (
       i++;
     }
 
+    if (previousVersionTime && previousVersionTime.getTime() >= resolutionNow.getTime()) {
```
brianorwhatever: maybe we should have a small allowance?

brianorwhatever (changes requested):
looks pretty good. I think the Z time issue needs to be fixed.

src/utils/iso8601-datetime.ts
rmlearney-digicatapult: Hi @brianorwhatever thanks for the review :) this is where I found the spec slightly confusing

The normative language is 

`
The versionTime MUST be a valid UTC [[ref: ISO8601]] string with explicit Z (or +00:00); values without a zone designator, or expressing a non-UTC zone, MUST be rejected
`

My read was that rejecting a non-UTC zone meant that a valid UTC zone was accepted (meaning `-12:00` to `+14:00`) but if the real read is **only** `+00:00` I'm happy to fix

@swcurran is this a correct interpretation of the normative language?

src/method_versions/method.v1.0.ts
rmlearney-digicatapult: yes I think the spec allows for up to 5 mins, will implement

src/utils/iso8601-datetime.ts
swcurran: Yes, that is both my read and understanding of the intention.

rmlearney-digicatapult: Have made the validation stricter so it now only accepts `Z` or `+00:00`

brianorwhatever (approved):

PR #121, opened 2026-06-11
fix(core): further spec alignment

rmlearney-digicatapult:
Fixes for the following deviations from spec:

* Request for valid earlier `versionId` or `versionTime` now return that version, even if later entries are invalid
* Request for invalid earlier `versionId` or `versionTime` now returns `notFound`
* Deactivated DIDs no longer return a `DIDDoc`, with metadata indicating deactivation
* Unspecified `nextKeyHashes` now retain their previous values, unless controller explicitly publishes an empty array
* Unspecified `updateKeys` now retain their prior values, unless controller explicitly publishes an empty array. While pre-rotation is active, `updateKeys` must always be specified explicitly
* Implicit services now inject fragment IDs (`#files`, `#whois`), but not absolute IDs (still spec compliant)
* `DIDDoc` state persisted with sparse `updateDID()` calls
* Requested DID must match a top-level `state.id` in a valid version
* `SCID` addition only allowed in first log
* Enforce `SHA256` SCID algorithm during resolution
* Enforce `did:key` body & fragment equality
* Portability `true/false` transition checks
* Permit version increase in log history
* Witness key must be `Ed25519`

Added regression tests

**NB**
Code paths fixed in `v1.0` not `v0.5` which has been left untouched

### DIFF
--- a/src/method_versions/method.v1.0.ts
+++ b/src/method_versions/method.v1.0.ts
@@ -66,6 +66,33 @@ const requireDidId = (id: string | undefined): string => {
   return id;
 };
 
+const parseAndValidateVersionId = (versionId: string, expectedVersionNumber: number) => {
+  const firstDashIndex = versionId.indexOf('-');
+  const lastDashIndex = versionId.lastIndexOf('-');
+
+  if (firstDashIndex === -1 || firstDashIndex !== lastDashIndex) {
+    throw new Error(`versionId '${versionId}' must contain exactly one '-' separator`);
+  }
+
+  const version = versionId.slice(0, firstDashIndex);
+  const entryHash = versionId.slice(firstDashIndex + 1);
+
+  if (!/^\d+$/.test(version)) {
+    throw new Error(`versionId '${versionId}' must have a numeric version prefix`);
+  }
+
+  if (entryHash.length === 0) {
+    throw new Error(`versionId '${versionId}' must have a non-empty hash component`);
+  }
+
+  const versionNumber = Number(version);
+  if (versionNumber !== expectedVersionNumber) {
+    throw new Error(`version '${version}' in log doesn't match expected '${expectedVersionNumber}'.`);
+  }
+
+  return { version, versionNumber, entryHash };
+};
+
 export const createDID = async (options: CreateDIDInterface): Promise<CreateDIDResult> => {
   if (!options.updateKeys) {
     throw new Error('Update keys not supplied');
@@ -243,11 +270,8 @@ export const resolveDIDFromLog = async (
   try {
     while (i < resolutionLog.length) {
       const { versionId, versionTime, parameters, state, proof } = resolutionLog[i];
-      const [version, entryHash] = versionId.split('-');
+      const { version, versionNumber, entryHash } = parseAndValidateVersionId(versionId, i + 1);
       const previousWitness = meta.witness ? deepClone(meta.witness) : undefined;
-      if (parseInt(version, 10) !== i + 1) {
-        throw new Error(`version '${version}' in log doesn't match expected '${i + 1}'.`);
-      }
       meta.versionId = versionId;
       if (!versionTime) {
         throw new Error(`version '${version}' is missing versionTime`);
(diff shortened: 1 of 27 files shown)


swcurran:
Confirm that the branch successfully runs the test-suite.

brianorwhatever:
ok ready to fix these conflicts and get this one in too 😃

rmlearney-digicatapult:
Could we release a new minor version after this merges?

swcurran:
> Could we release a new version after this merge?

Yes. What do you suggest the version be? 2.7.6? 2.8.0? 3.0.0?

rmlearney-digicatapult:
`2.8.0`, I'm thinking no breaking API changes for major version bump at this time

src/method_versions/method.v1.0.ts
```diff
@@ -578,7 +727,10 @@ export const updateDID = async (
   const signedProof = await options.signer.sign({ document: prelimEntry, proof: proofTemplate });
   const allProofs: DataIntegrityProof[] = [{ ...proofTemplate, proofValue: signedProof.proofValue }];
   prelimEntry.proof = allProofs;
-  const keysToVerify = lastMeta.prerotation ? params.updateKeys : lastMeta.updateKeys;
+  const keysToVerify = lastMeta.prerotation ? currentUpdateKeys : lastMeta.updateKeys;
```
brianorwhatever, line 735: When pre-rotation is active, updateDID requires updateKeys (line 624 guard) and verifies the new entry's proof against the new keys (keysToVerify = currentUpdateKeys, line 734) — so its own check always passes. But it never calls newKeysAreInNextKeys() to confirm the new keys were pre-committed in the prior entry's nextKeyHashes. Resolution does enforce this (lines 398-400). So a caller who
  rotates to keys not pre-committed gets a successful updateDID() whose log throws Invalid update key ... Not found in nextKeyHashes on every subsequent resolve. The validation should run in updateDID up front, not only at resolve time.

src/method_versions/method.v1.0.ts
```diff
@@ -41,6 +53,8 @@ import {
 } from '../utils/iso8601-datetime';
 import { countVerifiedWitnessApprovals, fetchWitnessProofs, validateWitnessParameter } from '../witness';
 
+const hasOwn = <K extends PropertyKey>(obj: object, key: K): obj is Record<K, unknown> => Object.hasOwn(obj, key);
+
 const VERSION = '1.0';
 const PROTOCOL = `did:${METHOD}:${VERSION}`;
```
brianorwhatever: I don't think these are used anymore

```suggestion
```

src/method_versions/method.v1.0.ts
```diff
@@ -183,19 +224,24 @@ export const createDID = async (options: CreateDIDInterface): Promise<CreateDIDR
 export const resolveDIDFromLog = async (
   log: DIDLog,
   options: ResolutionOptions & { witnessProofs?: WitnessProofFileEntry[] } = {}
-): Promise<{ did: string; doc: DIDDoc; meta: DIDResolutionMeta }> => {
+): Promise<{ did: string; doc: DIDDoc | null; meta: DIDResolutionMeta }> => {
```
brianorwhatever, line 228: I think technically this is a breaking API change as it changes the shape of the responses signature but I don't think it's a big deal so we should go with 2.8.0 like you said

brianorwhatever (commented):
This looks good just found the one issue. I think the PR I opened to your repo is still required as I don't think it's solved here. I am happy with the next release being 2.8.0 however lets get that portability fix in first as well

swcurran:
Confirmed that the latest commit works with the test-suite, and continues to fix all of the remaining failures in the ts generator.

rmlearney-digicatapult:
Have found a bug, converted to draft

rmlearney-digicatapult:
> This looks good just found the one issue. I think the PR I opened to your repo is still required as I don't think it's solved here. I am happy with the next release being 2.8.0 however lets get that portability fix in first as well

That PR looks good

src/method_versions/method.v1.0.ts
rmlearney-digicatapult: Should be fixed now

brianorwhatever (approved):
looks good!

PR #122, opened 2026-06-12
Add tests (buffer, hash, http, errors).

dmitrizagidulin:
Adds four new test files that cover low-level utility modules and the HTTP resolution path, none of which currently have dedicated tests.

### `test/buffer.test.ts`

Unit tests for `src/utils/buffer.ts` (`createBuffer`, `bufferToString`,
`concatBuffers`). The buffer utilities switch between Node's `Buffer` and web
APIs (`TextEncoder`/`atob`/`btoa`) based on `config.isBrowser`, and both
implementations must agree. The suite runs twice via `describe.each`, once per
branch, by toggling `config.isBrowser`: utf-8 round-trips, base64
decode/encode, hex encoding, and buffer concatenation.

### `test/crypto-utils.test.ts`

Unit tests for `src/utils/crypto.ts`: `createHash` digest length and
`createHashHex` against the known SHA-256 test vector for `'hello'`.

### `test/multiformats.test.ts`

Unit tests for `src/utils/multiformats.ts`:

- base58btc round-trips (including leading-zero-byte preservation) and
  rejection of characters outside the alphabet
- multibase round-trips for `base58btc` and `base64url` plus rejection of
  too-short strings and unsupported encoding prefixes
- multihash round-trips (raw and through multibase) plus rejection of wrong
  digest lengths, truncated varints, truncated digests, and unsupported
  algorithms

### `test/resolve-http.test.ts`

Coverage of the HTTPS resolution path with a stubbed global `fetch` (no
network access; restored after each test):

- `resolveDID` fetches the log from the `.well-known/did.jsonl` URL and
  resolves successfully
- error mapping to `DidResolutionError` and `problemDetails`: HTTP 404 and an
  empty log map to `notFound`; an invalid log, a SCID mismatch between the DID
  and its log, and a network failure map to `invalidDid`
- `fetchLogFromIdentifier` builds path-qualified URLs for path-based DIDs
- `fetchWitnessProofs` fetches `did-witness.json` alongside the log and
  returns an empty list when the file is missing or the fetch fails


rmlearney-digicatapult:
Thanks @dmitrizagidulin I agree with expanding the unit test base, particularly for self-rolled functions in `buffer.ts` and `multiformats.ts` but I don't think we need the `crypto-utils` test which is quite minimal and only acts to test the underlying library function from `@noble/hashes/sha2.js` which itself is well tested

dmitrizagidulin:
@rmlearney-digicatapult Understood, I agree. I removed `crypto-utils` test.

swcurran:
Confirm that the branch successfully runs the test-suite.  @rmlearney-digicatapult -- you know have the power to approve this PR.  And @dmitrizagidulin -- you have the power to approve other PRs in the repo.

brianorwhatever (approved):

PR #123, opened 2026-06-19
Fix host extraction for path-based did:webvh identifiers

brianorwhatever:
## Problem

Host extraction used `requireDidId(id).split(':').at(-1)`, which returns the **last** `:`-separated segment of a `did:webvh` identifier. That equals the host only when the DID has no path. For path-based DIDs (`did:webvh:{scid}:{domain}:{path}…`) it returns a path segment instead.

The most serious consequence is a hole in the **portability check** (`Cannot move DID: portability is disabled`): a non-portable DID with a path could be moved to a different domain undetected, as long as the trailing path segment matched — e.g. `…:example.com:dids:abc` → `…:evil.com:dids:abc` both reduce to `abc`.

This predates the recently-merged type-safety PR (#119), which only wrapped the existing `.at(-1)` logic — it didn't introduce it.

## Fix

- Add two helpers in `src/utils.ts` (both validate the `did:webvh:` prefix):
  - `getWebvhLocation(id)` → everything after the SCID (domain + path), for "has the DID moved?" comparisons.
  - `getWebvhHost(id)` → just the domain (`parts[3]`).
- `method.v1.0.ts` and `method.v0.5.ts`: portability/host comparisons now use `getWebvhLocation` (full location — domain *and* path), and the `domain` field uses `getWebvhHost`.

## Tests

The existing portability test used a path-less DID, so `.at(-1)` happened to return the right value and the bug was invisible; `paths.test.ts` only exercised `createDID`, never an update/resolve cycle.

Added a regression test — *DID log with portable false and a path should not resolve if domain moved* — that moves the domain while keeping the final path segment. Verified it **fails** against the old `.at(-1)` logic and **passes** with the fix.

Full suite: 185 pass / 0 fail. Build type-checks cleanly.

### DIFF
--- a/src/method_versions/method.v0.5.ts
+++ b/src/method_versions/method.v0.5.ts
@@ -21,6 +21,8 @@ import {
   deriveHash,
   findVerificationMethod,
   getBaseUrl,
+  getWebvhHost,
+  getWebvhLocation,
   parseCanonicalAddress,
   replaceValueInObject,
 } from '../utils';
@@ -197,7 +199,7 @@ export const resolveDIDFromLog = async (
       if (version === '1') {
         meta.created = versionTime;
         newDoc = state;
-        host = requireDidId(newDoc.id).split(':').at(-1) ?? '';
+        host = getWebvhLocation(requireDidId(newDoc.id));
         meta.scid = parameters.scid as string;
         meta.portable = parameters.portable ?? meta.portable;
         meta.updateKeys = parameters.updateKeys as string[];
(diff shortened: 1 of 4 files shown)


rmlearney-digicatapult:
This is already fixed in `fix/negative-tests` branch with `parseCanonicalAddress()` in `utils.ts`

brianorwhatever:
will take a look at that PR asap. looks like it's got some juice to it

PR #127, opened 2026-06-23
fix: rebuild controller from new location on portable move

### LINKED ISSUE #101: Portability does not work?
Hi,

so I tried to understand how did portability works with this library but I'm getting no satisfying results, it just seems to silently ignore the update.

@swcurran  I tried taking a look at your test suite, and while I implemented the description, it seems the output of the test is not testing the expected change. If you look at both resolutionResult.json or did.jsonl, they both expect or contain `example.com` instead of what I understand should be `example.org`: https://github.com/swcurran/didwebvh-test-suite/tree/main/vectors/portable-move

The only place I found something that would appear to be a portability action is here:

https://github.com/decentralized-identity/didwebvh-ts/blob/main/src/method_versions/method.v1.0.ts#L182, but that's in `resolveDidFromLog` and it feels a bit strange that the resolution should do the migration instead of an update call?

brianorwhatever:
updateDID ignored the domain/address option when deriving the new entry's identifier: createDIDDoc builds the document as `id: controller`, and the controller defaulted to the previous entry's state.id, so a requested move was silently dropped and the DID stayed at its old location.

Derive the controller from the requested location (address/domain/paths) while preserving the SCID, so a portable DID can actually move. Moving a non-portable DID now fails fast with a clear error instead of producing an entry that only fails later at resolution. Also accept `address` and `paths` on updateDID for parity with createDID.

Adds regression tests covering a successful portable move and rejection of a move on a non-portable DID.

Closes #101

### DIFF
--- a/test/portability.test.ts
+++ b/test/portability.test.ts
@@ -116,4 +116,40 @@ describe('Portability', () => {
       'does not match SCID in log'
     );
   });
+
+  test('Portable DID moves to a new domain via the domain option (#101)', async () => {
+    const updateResult = await updateDID({
+      log: portableDID.log,
+      domain: 'example.org',
+      signer: createTestSigner(authKey),
+      updateKeys: [authKey.publicKeyMultibase!],
+      verificationMethods: asPublicVerificationMethods(authKey),
+      verifier: testImplementation,
+    });
+
+    const originalDid = portableDID.log[0].state.id as string;
+    const scid = originalDid.split(':')[2];
+
+    // The new entry's state.id must reflect the new location, same SCID
+    expect(updateResult.log[1].state.id).toBe(`did:webvh:${scid}:example.org`);
+
+    // And it must resolve to the moved DID
+    const resolved = await resolveDIDFromLog(updateResult.log, { verifier: testImplementation });
+    expect(resolved.did).toBe(`did:webvh:${scid}:example.org`);
+    expect(resolved.doc?.id).toBe(`did:webvh:${scid}:example.org`);
+  });
+
+  test('Non-portable DID rejects a move to a new domain (#101)', async () => {
+    await expect(
+      updateDID({
+        log: nonPortableDID.log,
+        domain: 'example.org',
+        signer: createTestSigner(authKey),
+        updateKeys: [authKey.publicKeyMultibase!],
+        verificationMethods: asPublicVerificationMethods(authKey),
+        verifier: testImplementation,
+      })
+    ).rejects.toThrow('Cannot move DID: portability is disabled');
+  });
+
 });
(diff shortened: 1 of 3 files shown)


brianorwhatever:
fairly certain i fixed the lint issue and its still failing..

swcurran (approved):

rmlearney-digicatapult:
Have submitted a lint fix PR to your branch @brianorwhatever

test/portability.test.ts
```diff
@@ -116,4 +116,80 @@ describe('Portability', () => {
       'does not match SCID in log'
     );
   });
+
+  test('Portable DID moves to a new domain via the domain option (#101)', async () => {
```
rmlearney-digicatapult, line 120: don't think we need the #101 in the test name

test/portability.test.ts
```diff
@@ -116,4 +116,80 @@ describe('Portability', () => {
...
+    // The new entry's state.id must reflect the new location, same SCID
+    expect(updateResult.log[1].state.id).toBe(`did:webvh:${scid}:example.org`);
+
+    // And it must resolve to the moved DID
+    const resolved = await resolveDIDFromLog(updateResult.log, { verifier: testImplementation });
+    expect(resolved.did).toBe(`did:webvh:${scid}:example.org`);
+    expect(resolved.doc?.id).toBe(`did:webvh:${scid}:example.org`);
+  });
+
+  test('Non-portable DID rejects a move to a new domain (#101)', async () => {
```
rmlearney-digicatapult, line 142: remove #101

test/portability.test.ts
```diff
@@ -116,4 +116,80 @@ describe('Portability', () => {
...
+        domain: 'example.org',
+        signer: createTestSigner(authKey),
+        updateKeys: [authKey.publicKeyMultibase!],
+        verificationMethods: asPublicVerificationMethods(authKey),
+        verifier: testImplementation,
+      })
+    ).rejects.toThrow('Cannot move DID: portability is disabled');
+  });
+
+  test('Re-passing a bare domain on a pathed DID preserves the paths (#101)', async () => {
```
rmlearney-digicatapult, line 155: remove #101

test/portability.test.ts
```diff
@@ -116,4 +116,80 @@ describe('Portability', () => {
...
+      verificationMethods: asPublicVerificationMethods(authKey),
+      verifier: testImplementation,
+    });
+
+    expect(updateResult.log[1].state.id).toBe(pathedDID.did);
+    const resolved = await resolveDIDFromLog(updateResult.log, { verifier: testImplementation });
+    expect(resolved.did).toBe(pathedDID.did);
+  });
+
+  test('Portable DID moves to a pathed location via the address option (#101)', async () => {
```
rmlearney-digicatapult, line 180: #101

rmlearney-digicatapult (approved):
Some small changes to test names

swcurran:
Merging things with test errors is a bit painful, but did it.  Hope that lint issue wasn't. :-)

PR #132, opened 2026-06-24
fix(versionTime): trim ms, update tests

rmlearney-digicatapult:
Trimmed `versionTime` to second precision

This change was a lot more extensive than anticipated because previously in the runtime we automatically bumped every log creation by 1 msec to help the tests pass

I stripped this out so that creation uses the actual wall clock time

This meant:
* we needed a new test helper (`nextSecond`) to bump each log entry by 1 second in chained tests
* we needed to support `--updated` pass-through in the CLI (mirroring the API) and use this in tests
* use a delay in `deactivateDID` tests because we cannot pass through `versionTime` (`--updated`) to this method - I can add this as well if we want users to be able to set deactivation time

Using Bun's mock timers was much more ugly than the single `nextSecond` helper, and because the CLI e2e tests execute as a sub-process, these weren't available in the `cli-e2e` test suite

NB the test suite now takes about 3+ seconds longer to run which is annoying

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -45,6 +45,7 @@ Options:
   --domain [domain]         DEPRECATED: Use --address instead. Domain for the DID (backwards compatibility).
   --log [file]              Path to the DID log file (required for resolve, update, deactivate)
   --output [file]           Path to save the updated DID log (optional for create, update, deactivate)
+  --updated [timestamp]     Explicit update timestamp in UTC ISO8601 (e.g. 2026-01-01T12:00:00Z)
   --portable                Make the DID portable (optional for create)
   --witness [witness]       Add a witness (can be used multiple times)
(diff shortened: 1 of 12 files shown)


swcurran:
Looks good -- ran the test suite and this matches previous results and the Rust library processes the Logs.  Interesting Rust issue with the portable-move test that I'll look into.

@brianorwhatever need to get this one in before the release.

rmlearney-digicatapult:
@swcurran see if this PR resolves the suspected msec conflict with the rust implementation

swcurran:
Good details in the comment -- appreciated!  I wondered about the size of the change :-).

rmlearney-digicatapult:
You're very welcome :)

swcurran:
Rust working except for portable-move (which looks OK to me...at first glance, but don't really know) and the issue with the missing witness on witness-update.

brianorwhatever (approved):
looks good

PR #133, opened 2026-06-24
fix(portability): populate alsoKnownAs on move

rmlearney-digicatapult:
Fix rust interop test around `portable-move`

TS implementation wasn't populating `alsoKnownAs` with previous DID on move, now spec aligned with `DIFF` not `FAIL`

Remaining mismatch is in resolved output shape (implementation-specific, not spec misalignment):
* Rust returns absolute implicit service IDs like `did:webvh:...#files` and `did:webvh:...#whois`while TS expects `#files` and `#whois`.
* Rust includes extra metadata fields such as `deactivated`, `portable`, `scid`, `watchers` and `witness`
* Rust returns `didResolutionMetadata.contentType` where some expected outputs omit or differ on that shape.

### DIFF
--- a/src/method_versions/method.v1.0.ts
+++ b/src/method_versions/method.v1.0.ts
@@ -758,13 +758,13 @@ export const updateDID = async (
     doc.alsoKnownAs = options.alsoKnownAs;
   }
 
-    if (controller !== lastEntryDid) {
-      const aliases = Array.isArray(doc.alsoKnownAs) ? [...doc.alsoKnownAs] : [];
-      if (!aliases.includes(lastEntryDid)) {
-        aliases.push(lastEntryDid);
-      }
-      doc.alsoKnownAs = aliases;
+  if (controller !== lastEntryDid) {
+    const aliases = Array.isArray(doc.alsoKnownAs) ? [...doc.alsoKnownAs] : [];
+    if (!aliases.includes(lastEntryDid)) {
+      aliases.push(lastEntryDid);
     }
+    doc.alsoKnownAs = aliases;
+  }
 
   const logEntry: DIDLogEntry = {
     versionId: lastEntry.versionId,
--- a/src/method_versions/method.v1.0.ts
+++ b/src/method_versions/method.v1.0.ts
@@ -758,6 +758,14 @@ export const updateDID = async (
     doc.alsoKnownAs = options.alsoKnownAs;
   }
 
+    if (controller !== lastEntryDid) {
+      const aliases = Array.isArray(doc.alsoKnownAs) ? [...doc.alsoKnownAs] : [];
+      if (!aliases.includes(lastEntryDid)) {
+        aliases.push(lastEntryDid);
+      }
+      doc.alsoKnownAs = aliases;
+    }
+
   const logEntry: DIDLogEntry = {
     versionId: lastEntry.versionId,
     versionTime: createdDate,
--- a/test/portability.test.ts
+++ b/test/portability.test.ts
@@ -160,11 +160,13 @@ describe('Portability', () => {
 
     // The new entry's state.id must reflect the new location, same SCID
     expect(updateResult.log[1].state.id).toBe(`did:webvh:${scid}:example.org`);
+    expect(updateResult.log[1].state.alsoKnownAs).toContain(originalDid);
 
     // And it must resolve to the moved DID
     const resolved = await resolveDIDFromLog(updateResult.log, { verifier: testImplementation });
     expect(resolved.did).toBe(`did:webvh:${scid}:example.org`);
     expect(resolved.doc?.id).toBe(`did:webvh:${scid}:example.org`);
+    expect(resolved.doc?.alsoKnownAs).toContain(originalDid);
   });
 
   test('Non-portable DID rejects a move to a new domain', async () => {


rmlearney-digicatapult:
@swcurran @PatStLouis @brianorwhatever happy to take your steer on aligning the resolved output shape

swcurran:
Ran the test suite and confirm that this resolves the interop issue with the Rust library.  Thanks @rmlearney-digicatapult !

rmlearney-digicatapult:
Have discovered why the witnessing tests don't align and am working on a fix. It's a spec noncompliance issue in the TS implementation so probably also needs to be merged before we release a new version

src/method_versions/method.v1.0.ts
```diff
@@ -758,6 +758,14 @@ export const updateDID = async (
     doc.alsoKnownAs = options.alsoKnownAs;
   }
 
+  if (controller !== lastEntryDid) {
+    const aliases = Array.isArray(doc.alsoKnownAs) ? [...doc.alsoKnownAs] : [];
+    if (!aliases.includes(lastEntryDid)) {
+      aliases.push(lastEntryDid);
```
brianorwhatever, line 764: this list could get pretty big… It makes me hesitate. 
Is this only happening on portability moves? or are there other reasons we enter the `controller !== lastEntryDid` branch?

src/cli.ts
```diff
@@ -45,6 +45,7 @@ Options:
   --domain [domain]         DEPRECATED: Use --address instead. Domain for the DID (backwards compatibility).
   --log [file]              Path to the DID log file (required for resolve, update, deactivate)
   --output [file]           Path to save the updated DID log (optional for create, update, deactivate)
+  --updated [timestamp]     Explicit update timestamp in UTC ISO8601 (e.g. 2026-01-01T12:00:00Z)
```
brianorwhatever: this scares me a bit too even though it’s already a self attested timestamp it makes me 🤔

rmlearney-digicatapult: Are you able to revert PR #132 and we'll fix the other way by bumping each new log entry by +1 second on update without introducing any of these changes

src/method_versions/method.v1.0.ts
rmlearney-digicatapult: I'm guided by 2 sections of the normative language in the spec:

The [[ref: DID Log]] of the renamed DID MUST contain all of the [[ref: log entries]] from the creation of the DID.

The [[ref: DIDDoc]] MUST contain the prior DID string as an alsoKnownAs entry.

We _could_ trim the entry in the `DIDDoc` to just the most recent `alsoKnownAs` entry

rmlearney-digicatapult: The Rust implementation also preserves prior aliases by carrying forward state, then appends the just-previous DID meaning no trimming

rmlearney-digicatapult:
Going to rebase after #137 lands

src/method_versions/method.v1.0.ts
brianorwhatever: ok sounds like the thing to do then. lets match the rust impl here right @swcurran? not sure the spec considered this list could be a growth vector that would get exponentially worse for repeatedly moved DIDs.

rmlearney-digicatapult:
Now rebased

src/method_versions/method.v1.0.ts
swcurran: I think it should have the full list.  Anyone using this feature MUST be aware that this list is going to get long if it moves it every time, and the spec makes it clear.  We could put an added warning in the security (?) concerns, but its the DID Controller that is making a mess if that is how the use case is designed.

swcurran (approved):
LGTM and tested using the test suite.

brianorwhatever:
Ok so which of these 2 scenarios is correct (both are possible after this merge)

Case A — caller doesn't re-pass alsoKnownAs on each update. The move entry gets [prevDID], but the next regular update rebuilds a fresh doc with no alsoKnownAs, so the alias vanishes from subsequent entries. The alias lives in exactly one entry per move.

Case B — caller carries the document forward (fetch latest doc, modify, pass it back — the common and arguably correct pattern). Now alsoKnownAs accumulates and, because the log is append-only and every entry stores the full state document, the entire accumulated alias list is duplicated into every subsequent log entry, including key-rotation entries.

swcurran:
Sorry if I merged prematurely.  We thoughts on your comment:

I think Case B is correct.  Yes, painful, but that is what the spec says. That said, lots of things in the DIDDoc are repeated -- "id", "controller" at least.

Case A (if I understand it -- the controller handles all updates to the DIDDoc and doesn't put in the alsoKnownAs) is an update error caught by the implementation that refuses to publish a DID Log without a complete alsoKnownAs field.

I see you say that Case A is possible in what is implemented.  That would be bad IMHO.

Is the library putting in the alsoKnownAs "behind the back" of the DID Controller.  E.g. is it altering the DIDDoc it is given?  That could be OK if we agree to it, but would definitely want to make sure we don't do that if the controller has already added it.

brianorwhatever:
Ok sorry ya we're only working with Case B now. I was using dated mental models.

updateDID now does doc = deepClone(lastEntry.state) and only overrides fields when the option is explicitly passed, so alsoKnownAs persists and accumulates without the controller re-passing it (same as id/controller repeating, as you noted). Not behind the controller's back in a destructive sense: on a move it appends the prior DID with a dedup guard (!aliases.includes(lastEntryDid)), so it's a no-op if the controller already added it, and it never removes entries. The one thing we don't do is enforce completeness. We guarantee the immediate predecessor on a move and rely on the clone to carry earlier hops, but we don't reject an incomplete alsoKnownAs. If we want the "refuse to publish without a complete chain" behavior you described, that'd be a follow-up validation.

After talking to claude about it a bit I am not as concerned with the log growth though it is real. Especially if we consider these DIDs to be long live and we are pushing the portability piece.

brianorwhatever:
of note: claude mentioned gzip would compress this 80-90% 👍

PR #135, opened 2026-06-25
fix(witness): verify proofs against their own versionId; govern list changes by the previous list

brianorwhatever:
## Summary

Two spec-conformance bugs in witness handling, both surfaced by the [didwebvh-test-suite](https://github.com/decentralized-identity/didwebvh-test-suite) `witness-update` interop scenario. With these fixes, the resolver correctly handles every other implementation's witness files (rust, python, java, java-eecc, dart).

### Bug A — resolver verified witness proofs against the wrong `versionId`

`countVerifiedWitnessApprovals` hashed the **target** entry's `versionId` instead of the `did-witness.json` entry's own `versionId` (the value the witness actually signed). This defeats the spec's cumulative-approval rule and makes the resolver unable to verify a **pruned** witness file.

- Spec (§The Witness Proofs File): *"A valid proof from a witness carries the implication that ALL prior DID Log entries are also approved"*, and controllers **SHOULD** prune to *"keeping only the latest proof for each witness."*
- Spec (§Verifying Witness Proofs During Resolution, step 5): a target is met by proofs *"whose versionId matches the current or any later published entry."*
- Reproduced: resolving the suite's Rust `witness-update` vector (a pruned file) threw `Witness threshold not met`. Fix lives in `src/witness.ts` and is shared by both v1.0 and v0.5.

### Bug B — witness-list change applied to the entry that defines it

When an entry **replaced/reduced** the witness list, that entry was required to be witnessed by the **new** list. Per spec the new list becomes active only *after* the entry is published, so the entry must be witnessed by the **previous** list.

- Spec (§Witness Lists): *"If a DID log entry contains a new (replacement) list of witnesses … that new list becomes active AFTER the new DID log entry has been published."*
- Activation from `{}` (no witnesses → witnesses) remains immediately active, per §The `witness` Parameter — preserved.
- Fix reorders `getRequiredWitnessForEntry` in `method.v1.0.ts` (v1.0-only; v0.5 witness-change semantics not touched).

## Changes

- `src/witness.ts` — verify each proof against `proofSet.versionId` (covers v1.0 and v0.5).
- `src/method_versions/method.v1.0.ts` — previously-active list governs a list change/removal; activation-from-`{}` is the only immediate case.
- `test/witness.test.ts` — rewrote two tests that encoded the buggy behavior; added a pruned cumulative-approval test and a list-reduction rejection test.

## Testing

- Full suite green (**257 pass / 0 fail**); `bun run build` (tsc strict) clean.
- New/rewritten behavior tests confirmed RED before the fix and GREEN after, on this base.
- Interop verified against the official test-suite vectors: the resolver now resolves rust/python/java/java-eecc/dart `witness-update` and all `witness-threshold` vectors.

## Note for the test-suite

After Bug B, this library's **own** committed `witness-update` vector is now correctly rejected as under-witnessed (its reducing entry carries only one witness proof; every other implementation carries both). The TS vector should be regenerated in the test-suite repo so the reducing entry is witnessed by the full previous list.

### DIFF
--- a/src/method_versions/method.v1.0.ts
+++ b/src/method_versions/method.v1.0.ts
@@ -909,20 +909,14 @@ const getRequiredWitnessForEntry = (
 ): WitnessParameterResolution | undefined => {
   const explicitWitness = getEntryWitnessParameter(parameters);
 
-  if (explicitWitness !== undefined) {
-    if (isWitnessActive(currentWitness)) {
-      return deepClone(currentWitness);
-    }
-
-    if (isWitnessActive(previousWitness)) {
-      return deepClone(previousWitness);
-    }
-
-    return undefined;
-  }
-
+  // A list change takes effect only after its entry is published, so the previous list
+  // governs that entry; only activation from {} uses the new (current) list immediately.
   if (isWitnessActive(previousWitness)) {
     return deepClone(previousWitness);
+  }
+
+  if (explicitWitness !== undefined && isWitnessActive(currentWitness)) {
+    return deepClone(currentWitness);
   }
 
   return undefined;
--- a/src/witness.ts
+++ b/src/witness.ts
@@ -282,8 +282,9 @@ export async function countVerifiedWitnessApprovals(
 
         const { proofValue, ...proofWithoutValue } = proof;
 
-        // Create hashes
-        const canonicalizedData = canonicalizeStrict({ versionId: logEntry.versionId });
+        // Verify against the proof entry's own versionId (what the witness signed); a
+        // later proof cumulatively approves earlier entries.
+        const canonicalizedData = canonicalizeStrict({ versionId: proofSet.versionId });
         const canonicalizedProof = canonicalizeStrict(proofWithoutValue);
         const dataHash = await createHash(canonicalizedData);
         const proofHash = await createHash(canonicalizedProof);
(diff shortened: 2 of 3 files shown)


brianorwhatever:
yeah this is a pretty big bug I think. might be what you’re referencing [here](https://github.com/decentralized-identity/didwebvh-ts/pull/133#issuecomment-4794173717)?

rmlearney-digicatapult:
Yes spot on @brianorwhatever

brianorwhatever:
@rmlearney-digicatapult can you approve?  EDIT -> sorry didn’t mean to get pushy here I thought this was a different PR

rmlearney-digicatapult:
Afraid we still seem to be getting `FAIL witness-update/ts/resolutionResult.json (Witness proof threshold (2) was not met. Only (1) proofs were validated)` against `didwebvh-test-suite`

swcurran:
There is a second bug involved in the test-suite.  This PR fixes that the resolver in didwebvh-ts was not detecting the bad witness proof set. I have a PR to the test-suite that is properly generating the DID Log for the witness-update test to produce the correct witness proof set.  So we are safe to merge this one, and my test-suite PR will fix the fail.  I'm also going to add a negative test that use the bad DID Log / did-witness.json combination to make sure all did:webvh implementations detect the issue.

swcurran (approved):

PR #136, opened 2026-06-25
fix(versionTime): trim msec

rmlearney-digicatapult:
Smaller change based on `33e0fcb333f49700d11484a8cf19e816ef5c8426` (`HEAD` prior to merging PR #132 )

Use regex to trim trailing milliseconds, automatically bump each log creation time by 1 second when calling `createNextVersionTime()`

Not as clean as using wall clock time but:
* solves the rust issue
* minimises changes to the CLI (valid concern from @brianorwhatever)
* no changes to the test suite
* obviates the need for PR #134

### DIFF
--- a/src/utils/iso8601-datetime.ts
+++ b/src/utils/iso8601-datetime.ts
@@ -128,10 +128,11 @@ export function createNextVersionTime(
     return formatDate(requestedVersionTime);
   }
 
-  const now = new Date();
-  if (now.getTime() <= previous.getTime()) {
-    return formatDate(new Date(previous.getTime() + 1));
+  const nowFormatted = formatDate(new Date());
+  const nowTrimmed = new Date(nowFormatted);
+  if (nowTrimmed.getTime() <= previous.getTime()) {
+    return formatDate(new Date(previous.getTime() + 1000));
   }
 
-  return formatDate(now);
+  return nowFormatted;
 }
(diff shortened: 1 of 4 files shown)


src/utils/iso8601-datetime.ts
```diff
@@ -128,10 +128,11 @@ export function createNextVersionTime(
     return formatDate(requestedVersionTime);
```
brianorwhatever: Now that the generated path trims to second precision, this branch can pass an entry through that ends up **equal** to the previous `versionTime`: the check on the line above compares the *untrimmed* requested time, but the value returned here is *trimmed*. E.g. `previous = 2025-01-01T00:00:05Z`, requested `2025-01-01T00:00:05.400Z` → the untrimmed check passes (5400ms > 5000ms), then this returns `2025-01-01T00:00:05Z`, equal to `previous`, which then fails resolution.

Re-validate the trimmed value before returning (and the untrimmed `if (requested.getTime() <= previous.getTime())` check on the two lines above becomes redundant — safe to drop):

```suggestion
    const formatted = formatDate(requestedVersionTime);
    if (new Date(formatted).getTime() <= previous.getTime()) {
      throw new Error('versionTime must be greater than previous versionTime');
    }
    return formatted;
```

brianorwhatever: looks like you’re still working on this as I reviewed so this might be outdated already

rmlearney-digicatapult:
@brianorwhatever sorry, finished now. Ready for review

swcurran:
Looks like this is missing the recently merged witness handling, as it is now failing the new "negative" test for that.  I had expected to see a "Update PR to latest" button, but I don't see that.  So perhaps another rebase is needed.

Other than that, looks good and does what is expected.

rmlearney-digicatapult:
@swcurran merged manually

swcurran (approved):

Issue #130, opened 2026-06-24
Query parameter support


rmlearney-digicatapult:
At present in `resolveDIDFromLog` and `resolveDID` we support the following query parameters in `options`:

**MUST** from spec:
* `versionId`
* `versionTime`

**SHOULD** from spec:
* `versionNumber`

**Implementation specific**
* `verificationMethod`

All of these error with `notFound` or `invalidDID` as appropriate

Questions:

1. Should we remove support for the implementation-specific `verificationMethod` parameter? Does this present a security/privacy risk?
2. How should we handle combinations of query parameters - reject all combinations, or require satisfaction of all parameters before returning

swcurran:
What does the "verificationMethod" parameter do?  I had not heard of that one.

I think we should reject combinations (I assume by that you mean versionTime and versionId in a single call).  In theory we could wait to see if they are consistent, but that is far removed from parameter checking, complicated, and we should just discourage the practice immediately.

brianorwhatever:
verificationMethod returns the last state of the doc with that VM. I think it's a bad idea as it returns a historical doc without informing the client. I've removed it in #138


I also enforced only allowing a single parameter in that PR

swcurran:
I hope you mean just a single versionTime/versionId/versionNumber parameter vs. a single parameter in general.  Or at least in the future if we support more types of parameters, we should allow multiple -- just not multiple that target a specific version.

rmlearney-digicatapult:
Yes I can see that #138 ensures that either `versionTime` or `versionId` or `versionNumber` is passed but not a combination. Stripping `VM` is also a good move.

swcurran:
The `verifcationMethod` parameter is interesting.  I don't like the name, but perhaps `versionVM` might be reasonable. We could register that in the DID Extensions registry to make it common -- and reference it in the did:webvh spec.  I do think we need to give tools to enable access to historical data -- although "versionTime" is the most useful/powerful.

Not sure what you mean about returning an historical version without informing the client.  If the client knowingly put in a parameter that requested an historical version, is that not valid?  I can see providing other details in the DIDDoc Metadata to inform the caller that they have received an historical DID -- versionTime (in the DIDDoc), expiryTime (versionTime of the next Log Entry, if any), and perhaps some current state data.  We would want to check what is already being used by other DID Methods.

brianorwhatever:
[...]

PR #138, opened 2026-06-25
feat!: did-resolver interface + spec-conformant query parameters (#89, #130)

### LINKED ISSUE #89: Support did-resolver interface
Hi, 

I noticed that this package uses as a standalone `resolveDID()` API rather than implementing the [`did-resolver`](https://github.com/decentralized-identity/did-resolver) interface that other DID methods (like `did:web`, `did:ethr`) use.

Are there any technical reasons preventing compatability or was this an intentional design choice?

Thanks in advance for your insights!

brianorwhatever:
## Summary

Adopts the standard W3C [`did-resolver`](https://github.com/decentralized-identity/did-resolver) interface and settles the open query-parameter questions. **Breaking change — major version bump to 3.0.0.**

Resolves #89 (support did-resolver interface) and #130 (query parameter support).

## What changed

1. **`resolveDID` / `resolveDIDFromLog` return the standard resolution shape** `{ didResolutionMetadata, didDocument, didDocumentMetadata }` — a clean break from `{ did, doc, meta, controlled }`. They no longer throw on failure; errors come back as a result with `didDocument: null` and a standard `error` code.
2. **New `getResolver(config?)`** returns `{ webvh: DIDResolver }`, registrable in a `did-resolver` `Resolver` — `did:webvh` becomes a drop-in alongside `did:web`, `did:ethr`, etc. Works zero-config via a built-in **`defaultVerifier`** (Ed25519), overridable with `getResolver({ verifier })`.
3. **`verificationMethod` resolution removed entirely** — implementation-specific, not in any spec, and a footgun (returned a historical document version).
4. **Resolution selectors are mutually exclusive** — at most one of `versionId` / `versionTime` / `versionNumber`; any combination returns `invalidDidUrl`.
5. Both v1.0 and v0.5 map to the standard shape via a single mapping layer; internal `updateDID`/`deactivateDID` keep using the rich per-version core resolver unchanged.

## New modules

- `src/verifier.ts` — `defaultVerifier` (Ed25519 via `@stablelib/ed25519`).
- `src/resolver-result.ts` — pure mapping (`toResolutionResult`, `toErrorResult`, `assertSingleVersionSelector`, `mapErrorToCode`, `InvalidDidUrlError`) + result types.
- `src/resolver.ts` — `getResolver()`; parses DID-URL query params, enforces selector rules, returns `DIDResolutionResult`.

## Error model

Resolution failures surface on `didResolutionMetadata.error`: `notFound` (404 / empty log), `invalidDidUrl` (bad/duplicate selectors), `internalError` (non-404 HTTP, network/transport failures), and `invalidDid` (document-validation failures). RFC9457 `problemDetails` and the `controlled` extension ride along.

## Implementation note

`did-resolver` exposes `?`-query parameters as the raw, undecoded `parsed.query` string (only `;`-matrix params populate `parsed.params`). Since the did:webvh spec uses `?` query parameters, `getResolver` parses `parsed.query` itself, decoding per RFC 3986 (`decodeURIComponent`, `+` is literal) rather than `URLSearchParams` (form-encoding) so a `versionTime` with a `+HH:MM` offset isn't corrupted.

## Testing

- New unit tests: `defaultVerifier`, the mapping layer, mutual-exclusivity guard, error-code classification.
- New integration test: `getResolver()` in a real `did-resolver` `Resolver` — fresh DID, `?versionId`/`?versionNumber`/`?versionTime`, invalid selectors → `invalidDidUrl`, not-found, deactivated, zero-config.
- Migrated the full existing suite to the standard shape; deleted the obsolete `verificationMethod` resolution tests. 279 tests pass.

## Gates

`bun test` (279 pass), `bunx biome ci .`, `bun run check`, and `bun run build` all green.

## Notes / deferred follow-ups

- Rebased onto current `main` (incorporates the #132 versionTime-trim revert).
- Path/fragment **dereferencing** is intentionally out of scope: `resolve()` returns the DID document; dereferencing a DID URL to a sub-resource is a separate operation, consistent with `did:web`/`did:ethr`.
- A typed-error taxonomy (to perfectly distinguish a throwing custom verifier from an invalid document) is noted as a larger follow-up.


brianorwhatever:
this is a breaking change but I think it’s well worth it. I’m sure all of our other tinkering with things will have broken some DIDs anyways so I think a major version is in order anyways

src/resolver.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  const resolve: DIDResolver = async (
+    _did: string,
+    parsed: ParsedDID,
+    _resolver: Resolvable,
+    _options
+  ): Promise<DIDResolutionResult> => {
+    // did:webvh selectors arrive as DID-URL query parameters (`?versionId=`),
+    // which did-resolver exposes as the raw, undecoded `parsed.query` string.
+    // Matrix-style DID parameters (`;key=value`) land in `parsed.params`; accept
```
dmitrizagidulin: > Matrix-style DID parameters (`;key=value`) land in `parsed.params`; accept

Are these actually used anywhere? (Or are still in the spec?)

rmlearney-digicatapult:
One Google note on `@stablelib/ed25519` versus `node:crypto`:

`the standard verification function in @stablelib/ed25519 lacks the CFRG-recommended S-component group order check, which leaves it vulnerable to signature malleability. For purely browser-based Ed25519 cryptography, modern web applications standardly default to the native Web Crypto API (globalThis.crypto) or @noble/ed25519 for a hardened, pure-JS alternative`

examples/express-resolver.ts
```diff
@@ -41,6 +41,23 @@ const expressVerifier = new ExpressVerifier('key-prod-1', 'did:example:123#key-1
...
 const port = process.env.PORT ? parseInt(process.env.PORT, 10) : 8000;
 
+// Map a DID resolution error code to an HTTP status so callers can tell a
+// missing identifier (404) from a malformed request (400) or a resolver
+// failure (500).
+const getStatusCodeFromError = (errorType?: string): number => {
+  switch (errorType) {
+    case 'notFound':
+      return 404;
+    case 'invalidDid':
```
rmlearney-digicatapult, line 49: Same comment about fall-through but style is style :)

src/index.ts
```diff
@@ -1,8 +1,13 @@
 export { AbstractCrypto, createDocumentSigner, createProof, createSigner, prepareDataForSigning } from './cryptography';
 export * from './interfaces';
 export { createDID, deactivateDID, resolveDID, resolveDIDFromLog, updateDID } from './method';
+export type { GetResolverConfig } from './resolver';
+export { getResolver } from './resolver';
+export type { WebvhDocumentMetadata, WebvhErrorCode, WebvhResolutionMetadata } from './resolver-result';
+export { InvalidDidUrlError } from './resolver-result';
 export { deriveNextKeyHash, generateParallelDidWeb, parseDidKeyDid, parseDidKeyVerificationMethod } from './utils';
 export { MultibaseEncoding, multibaseDecode, multibaseEncode } from './utils/multiformats';
+export { defaultVerifier } from './verifier';
```
rmlearney-digicatapult, line 10: Do we want to expose this to external consumers? both `getResolver()` and `resolveDID` already allow for overrides

examples/elysia-resolver.ts
```diff
@@ -47,22 +47,13 @@ const WELL_KNOWN_ALLOW_LIST = ['did.jsonl'];
 // Helper function to map DID resolution errors to HTTP status codes
 const getStatusCodeFromError = (errorType?: string): number => {
   switch (errorType) {
-    case 'INVALID_DID':
-    case 'INVALID_DID_URL':
-    case 'INVALID_OPTIONS':
+    case 'invalidDid':
```
rmlearney-digicatapult, line 50: I prefer no fall-through in switch-case statements. I know it makes code more verbose but also easier to reason about and move around/update/etc. but everyone has a style :)

package.json
```diff
@@ -47,7 +46,9 @@
   },
   "dependencies": {
     "@noble/hashes": "^2.2.0",
+    "@stablelib/ed25519": "^2.1.0",
```
rmlearney-digicatapult: Could we swap for `@noble/curves` which is better audited, supported, more frequently downloaded, and in line with the already imported `@noble/hashes`?

src/method.ts
```diff
@@ -62,49 +64,29 @@ export const createDID = async (options: CreateDIDInterface): Promise<CreateDIDR
 export const resolveDID = async (
   did: string,
   options: ResolutionOptions & { witnessProofs?: WitnessProofFileEntry[] } = {}
-) => {
+): Promise<DIDResolutionResult> => {
   const activeDIDs = await getActiveDIDs();
   const controlled = activeDIDs.includes(did);
+  const verifier = options.verifier ?? defaultVerifier;
```
rmlearney-digicatapult, line 69: what alternative verifiers do you have in mind?

src/resolver-result.ts
```diff
@@ -0,0 +1,170 @@
+import type { DIDDocumentMetadata, DIDResolutionMetadata, DIDResolutionResult } from 'did-resolver';
+import type { DIDDoc, DIDResolutionMeta, ProblemDetails } from './interfaces';
+import { DidResolutionError } from './interfaces';
+
+export type WebvhErrorCode =
```
rmlearney-digicatapult: Re-declares string literals which already exist in `DidResolutionError` in `interfaces.ts` - can we unify?

src/resolver-result.ts
```diff
@@ -0,0 +1,170 @@
...
+    if (extras.controlled !== undefined) {
+      didResolutionMetadata.controlled = extras.controlled;
+    }
+    // Preserve the resolved document when the core produced one. A valid
+    // earlier version can be returned alongside a warning-level error (e.g. an
+    // explicit version selector that resolves cleanly while a later log entry
+    // fails witness verification); dropping it would hide a legitimate result.
+    return {
+      didResolutionMetadata,
+      didDocument: (core.doc as unknown as DIDResolutionResult['didDocument']) ?? null,
```
rmlearney-digicatapult: I don't like passing through `unknown` in runtime code for type safety. We should know the shape of the resolved object already

src/resolver-result.ts
```diff
@@ -0,0 +1,170 @@
...
+    };
+  }
+
+  const didResolutionMetadata: WebvhResolutionMetadata = { contentType: CONTENT_TYPE };
+  if (extras.controlled !== undefined) {
+    didResolutionMetadata.controlled = extras.controlled;
+  }
+  return {
+    didResolutionMetadata,
+    didDocument: (core.doc as unknown as DIDResolutionResult['didDocument']) ?? null,
```
rmlearney-digicatapult: Same comment about passing through `unknown` in runtime code

src/resolver.ts
```diff
@@ -0,0 +1,83 @@
...
+import { defaultVerifier } from './verifier';
+
+export interface GetResolverConfig {
+  verifier?: Verifier;
+}
+
+/**
+ * Returns a `did-resolver` registry entry for `did:webvh`, registrable in a
+ * `Resolver` alongside other DID methods. Works zero-config via the built-in
+ * Ed25519 verifier; pass `{ verifier }` to override.
```
rmlearney-digicatapult, line 14: What alternative verifiers do you have in mind?

src/resolver.ts
rmlearney-digicatapult: I don't think matrix params are in the DID spec

src/utils.ts
```diff
@@ -921,11 +921,11 @@ export const resolveVM = async (vm: string) => {
...
         .split('\n')
         .map((l) => JSON.parse(l));
-      const { doc } = await resolveDIDFromLog(logEntries, { verificationMethod: vm });
-      if (!doc) {
+      const { didDocument } = await resolveDIDFromLog(logEntries, {});
+      if (!didDocument) {
         throw new Error(`Verification method ${vm} not found`);
       }
-      return findVerificationMethod(doc, vm);
+      return findVerificationMethod(didDocument as unknown as DIDDoc, vm);
```
rmlearney-digicatapult: I don't like passing through `unknown` in runtime code for type safety. We should know the shape of the return object

rmlearney-digicatapult (changes requested):

src/method.ts
rmlearney-digicatapult: Ignore

src/resolver.ts
rmlearney-digicatapult: Ignore

package.json
dmitrizagidulin: +1 to `@noble/curves` rec.

swcurran:
FYI -- created an update to the test-suite to compensate for the API changes and things looked good. This PR is behind main so recent fixes are missing, but once those get added, the test-suite passes.

src/method.ts
```diff
@@ -118,16 +100,21 @@ export const resolveDID = async (
...
+): Promise<DIDResolutionResult> => {
+  const verifier = options.verifier ?? defaultVerifier;
+  try {
+    assertSingleVersionSelector(options);
+    const version = getWebvhVersionFromLog(log);
+    const result =
+      version === '0.5'
+        ? await v0_5.resolveDIDFromLog(log, { ...options, verifier })
+        : await v1.resolveDIDFromLog(log, { ...options, verifier });
     maybeWriteTestLog(result.did, log);
```
rmlearney-digicatapult: CLI now owns persistence

package.json
brianorwhatever: Done in 192fa9e — swapped everywhere (verifier, CLI, test utils, all three examples), with `zip215: false` so verification matches didwebvh-py (aries-askar → ed25519-dalek `verify_strict`). Rationale in the commit message.

src/resolver-result.ts
brianorwhatever: Done in 3163f59 — single `DidResolutionError` union in `interfaces.ts`; `WebvhErrorCode` and the enum are gone.

src/resolver-result.ts
brianorwhatever: Removed in 3163f59.

src/resolver-result.ts
brianorwhatever: Removed in 3163f59.

src/utils.ts
brianorwhatever: Removed in af200eb.

src/index.ts
brianorwhatever: Keeping it, deliberately: unlike did:web/did:key/did:ethr, did:webvh has to verify Data Integrity proofs at resolve time, so `getResolver()` needs a working default to be a zero-config drop-in — and the README documents that behavior. The closest ecosystem precedent (digitalbazaar's DID drivers) treats crypto suites as public, importable, pluggable artifacts; a hidden default with injectable overrides would be the odd shape. Consumers who want their own still pass `{ verifier }`.

src/resolver.ts
brianorwhatever: Settled: they're not in the DID spec — only did-resolver's legacy parser surfaces them. The resolver now rejects matrix-form *version selectors* with `invalidOptions` (silently ignoring them would resolve latest when a specific version was requested) and ignores any other matrix params, same posture as unknown query params.

examples/elysia-resolver.ts
brianorwhatever: Done — no fall-through in either example.

examples/express-resolver.ts
brianorwhatever: Done — no fall-through in either example.

src/method.ts
brianorwhatever: Yep — the rebase onto main picked up the `src/cli` split; the CLI-side changes all live in `src/cli/index.ts` now.

brianorwhatever:
Updated: rebased onto current main (conflicts resolved, the `src/cli` split absorbed) and addressed the review — @noble/curves migration with strict RFC 8032 verification, unified error model with `problemDetails.type` drawn from the [did:webvh resolution-error registry](https://didwebvh.info/latest/resolution-errors/), matrix-selector rejection, cast cleanups. Details in the inline replies and commit messages (3163f59, 192fa9e, af200eb). Much of this originates from @rmlearney-digicatapult's aviarytech#5 — thanks! — with two researched divergences noted there.

Question for the group: should this merge to `main`, or do we want a dev/release branch? It's breaking (3.0.0), #149 is queued behind it, and the #150–153 v0.5-removal cluster will be breaking too — batching them on a `v3.x` branch and cutting main over once the set is ready might beat landing sequential breaking changes on main. Happy either way.

rmlearney-digicatapult (approved):
Nice 👍

PR #140, opened 2026-06-27
chore(examples): update examples

rmlearney-digicatapult:
Bump dependencies in `examples` folder

Remove unnecessary example dependencies from main `devDependencies`

Regenerate both `bun.lock` files

Update `express-resolver` to `express` `v5` semantics


swcurran (approved):
LG/Safe to me

PR #141, opened 2026-06-27
chore(core): remove deprecated methods and enums

rmlearney-digicatapult:
Built on #140 

Remove deprecated `createSigner()` method

Remove deprecated `domain` parameter from `method v1.0` interface & runtime, use `address` instead

Special handling in `method v0.5` - type extension for `domain` and error if both `address` + `domain` or neither are provided

Update tests to use `address`

Remove legacy error enums


src/method_versions/method.v1.0.ts
```diff
@@ -101,9 +101,9 @@ export const createDID = async (options: CreateDIDInterface): Promise<CreateDIDR
   }
 
   // Parse address input with strict validation
-  const addressInput = options.address || options.domain;
+  const addressInput = options.address;
   if (!addressInput) {
-    throw new Error('Either address or domain must be provided');
+    throw new Error('address must be provided');
```
dmitrizagidulin:
```suggestion
    throw new Error('Address must be provided');
```

dmitrizagidulin (approved):
👍 
(minor error capitalization change request) looks good.

Issue #142, opened 2026-06-28
Fix unguarded Buffer usage (web bundler support)


dmitrizagidulin:
## Problem

`encodeBase64Url()` in `src/utils/multiformats.ts` uses Node's `Buffer` unconditionally:

```ts
// src/utils/multiformats.ts
export function encodeBase64Url(bytes: Uint8Array): string {
  const base64 = Buffer.from(bytes).toString('base64'); // <-- Buffer is not defined in the browser
  return base64.replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/, '');
}
```

It's the only unguarded `Buffer` reference in `src/`, and it's reachable from the public `multibaseEncode()` (whose default encoding is `BASE64URL_NO_PAD`). Bundlers like Vite/Rollup do not polyfill Node built-ins, so calling `multibaseEncode(bytes)` in a browser app throws `ReferenceError: Buffer is not defined`.

## Secondary issue: environment detection misroutes Web Workers

The `Buffer` polyfill in `src/utils/buffer.ts` gates on `config.isBrowser`, which is `typeof window !== 'undefined'` (`src/config.ts`). A Web/Service Worker has neither `window` nor `process`, so it's classified as "not browser" and falls into the `Buffer.from(...)` branch -- which also throws there.

## Recommended fix

1. Route base64/base64url through [`@scure/base`](https://github.com/paulmillr/scure-base)
   (`base64` / `base64urlnopad`) -- pure-JS, browser/RN-safe, and same
   maintainer family as the existing `@noble/hashes` dependency:

   ```ts
   import { base64urlnopad } from '@scure/base';

   export function encodeBase64Url(bytes: Uint8Array): string {
     return base64urlnopad.encode(bytes);
   }
   function decodeBase64Url(str: string): Uint8Array {
     return base64urlnopad.decode(str);
   }
   ```

   The `base64` branch of `createBuffer` / `bufferToString` in
   `src/utils/buffer.ts` can be unified onto `base64.decode` / `base64.encode`
   the same way.

2. For the remaining (hex/utf8) `Buffer` fast path in `buffer.ts`, gate on the
   capability directly instead of sniffing for `window`, so every non-Node
   runtime (browser, RN, Web/Service Workers) takes the pure-JS path:

   ```ts
   const hasNodeBuffer = typeof Buffer !== 'undefined';
   // ...use `if (hasNodeBuffer)` instead of `if (!config.isBrowser)`
   ```

PR #143, opened 2026-06-28
fix: Route base64 through scure/base to fix Buffer is not defined in browsers

dmitrizagidulin:
Addresses issue https://github.com/decentralized-identity/didwebvh-ts/issues/142


dmitrizagidulin:
Apologies, I don't have Bun installed; it looks like CI is failing because of the lockfile.

rmlearney-digicatapult:
@dmitrizagidulin please delete `package-lock.json` and regenerate `bun.lock` and we should be good to go!

Can you add `package-lock.json` to `.gitignore` as well please :)

EDIT: One additional thought - should we use `@scure/base` for everything and drop node's `Buffer`?

src/utils/buffer.ts
```diff
@@ -1,4 +1,10 @@
-import { config } from '../config';
+import { base64 } from '@scure/base';
+
+// Whether Node's Buffer is available in the current runtime. Gating on the
```
rmlearney-digicatapult: should we remove support for node's `Buffer` and route everything through `@scure/base` like Credo-ts does in `TypedArrayEncoder`?

rmlearney-digicatapult:
I've just explored a more extensive clean-up. It turns out that the majority of `buffer.ts` is only used for `env` conversion/validation. Only `concatBuffers` is runtime-critical for both library and CLI. 

It's possible to use `@scure/base` directly within `config.ts` and primitive `TextDecoder/TextEncoder` to strip `buffer.ts` right down to the single method.

It's also possible to use `@scure/base` within `multiformats.ts` to significantly reduce all of the base conversion and validation logic.

I'll try to submit a PR to your branch @dmitrizagidulin for your review

rmlearney-digicatapult:
@dmitrizagidulin please merge #144 into this branch if you approve :)

dmitrizagidulin:
> @dmitrizagidulin please merge #144 into this branch if you approve :)

Merged, thank you so much!

rmlearney-digicatapult (approved):

PR #145, opened 2026-06-30
fix(dids): align service IDs with other implementations

rmlearney-digicatapult:
After running `didwebvh-test-suite` against `main` there are 2 main areas of `diff`, now resolved.

* Service ID format: Change from relative `#files` to absolute `did:webvh:SCID:example.com#files`
* Service endpoint: Add trailing slash to match other implementations
* Implicit services:  Properly formatted with absolute IDs


dmitrizagidulin (approved):
👍 
(though, for the record, I disagree with the 'default services injection' approach, and removed it in my fork)

rmlearney-digicatapult:
Thanks @dmitrizagidulin !

I'm afraid service injection is a MUST in the spec

If the "implicit" did:webvh services (as defined in the DID URL Resolution section) are not already present in the [[ref: DIDDoc]], they MUST be added. These services are the relativeRef service with id: "#files" or id: "<did>#files" and the whois service with id: "#whois" or id: "<did>#whois", with the serviceEndpoint for both derived from the DID-to-HTTPS transformation

PR #146, opened 2026-07-07
fix(updateDID): make timestamp optional on method

rmlearney-digicatapult:
The updateDID method currently exposes `updated` as a parameter. This should be optional so that normal callers do not think they can provide a timestamp.

Even if a caller passes the optional timestamp it still passes through internal validation for `iso8601` compliance and clock skew.


dmitrizagidulin (approved):
👍 looks good

PR #156, opened 2026-07-13
chore(tooling): migrate from bun to node

rmlearney-digicatapult:
Migrate from `bun` to `node` + `TS 7` + `vitest`


swcurran:
I don't see an update to README.md, which I think would be pretty important with this one :-).

brianorwhatever (approved):
You will be missed bun ❤️

PR #157, opened 2026-07-14
refactor(core): refactor method.v1.0.ts

rmlearney-digicatapult:
Significant refactoring with _very minimal changes_ to the existing test suites, mostly additions.

Tested against `didwebvh-test-suite` which passes all positive and negative tests. DIFF count remains high (also cross-suite so TS not unique) but the same DIFFs as `main`

`method.v1.0.ts` was composed of large monolithic functions, `utils.ts` was ballooning with a mix of one-line helpers and large functions

**NB** this is the first stage of refactoring out `method.v0.5.ts` so the naming of `method.v1.0.entries.ts` and `method.v1.0.resolution.ts` will change after further refactoring 

* centralised all DID Document helpers into new `did-document.ts`
* moved dateTime functionality into `iso8601-datetime.ts`
* moved hashing into `crypto.ts`
* moved constants into `constants.ts`

`resolveV1Log()` now acts as a helper orchestrator to progress through log entry validation using the methods inside `method.v1.0.resolution.ts`
* initialised a shared `resolverContext` object for processing through additional stages
* loop through & process each sequential log entry, differentiating between genesis and non-genesis entries
* enforce witness requirements, add default services if missing
* preserve captured results if later entries unprocessable/invalid
* final checks and result shaping including selector handling

`utils.ts` now contains mainly DID & URL parsing functionality, object manipulation, and verificationMethod-related functionality. There is still some leftover disk I/O which will be removed in future. Also reduced duplicated logic where possible.

Implemented further tests to increase codebase coverage to >90% across the board in areas that will be kept (not disk I/O, `config.ts`, `method.v0.5.ts`), with a `test:coverage` package script


swcurran:
Ran the test suite and noticed that all of the DID Logs changed and dug in a bit. The only content change I can see is that DIDDocs being produced includes the following:

```json
    "assertionMethod": [],
    "keyAgreement": [],
    "capabilityDelegation": [],
    "capabilityInvocation": [],
```

Since the DIDDoc is the domain of the controller (did:webvh doesn't have an opinion), it's not a did:webvh issue, but I'm curious as to why it is added, and the `[]` values for each.  Is that content coming from the TS library, or from the inputs into the TS library?

rmlearney-digicatapult:
@swcurran that should be cleaned up now and covered by a test

src/did-document.ts
```diff
@@ -0,0 +1,271 @@
...
+  const vms = verificationMethod.map((vm) => ({
+    ...vm,
+    id: vm.id ?? createVMID(vm, did),
+    controller: vm.controller ?? did ?? undefined,
+  }));
+  all.verificationMethod = vms;
+
+  for (const vm of vms) {
+    const relationship = vm.purpose;
+    if (!relationship) {
```
brianorwhatever, line 128: `main` treated a missing `purpose` as `authentication`: `all.authentication = vms.filter(vm => !vm.purpose || vm.purpose === 'authentication')`. This loop `continue`s instead, so a VM with no `purpose` lands in `verificationMethod` and in no relationship at all — the DID can't authenticate with it. Separate from the empty-array cleanup @swcurran spotted, and it affects `method.v0.5.ts` too now that it shares `createDIDDoc`. `test/did-document.test.ts:429` pins the new behaviour, so I assume it's deliberate — worth confirming and noting in the description.

Worth calling out the win in the same rewrite: `main` recomputed relationship ids via a second `createVMID()` call, so a VM without `publicKeyMultibase` got a different random suffix than its own `verificationMethod.id`. Single-pass fixes that.

src/index.ts
```diff
@@ -1,11 +1,19 @@
-export { AbstractCrypto, createDocumentSigner, createProof, prepareDataForSigning } from './cryptography';
+export {
+  AbstractCrypto,
+  createDataIntegrityProofTemplate,
+  createDocumentSigner,
+  prepareDataForSigning,
+  signDataIntegrityProof,
+} from './cryptography';
```
brianorwhatever, line 7: Three breaking changes land under a `chore:` title: `createProof` is gone from the public exports here, `controller` is dropped from `CreateDIDInterface`/`UpdateDIDInterface` (and `updateDID`'s `if (options.controller)` branch removed in both v1.0 and v0.5 — dead in create, but live in update), and `README.md:285` still documents `createProof`. Could this be `refactor!:` with a `BREAKING CHANGE:` footer, plus a README update?

src/witness.ts
```diff
@@ -58,41 +35,25 @@ export async function createWitnessProof(
...
+  const proofTemplate = createDataIntegrityProofTemplate({
     verificationMethod,
     created,
     proofPurpose: 'assertionMethod',
+  });
+
+  const adaptedSigner: Signer<{ versionId: string }> = {
+    getVerificationMethodId: () => verificationMethod,
+    sign: async ({ document, proof }): Promise<{ proofValue: string }> => {
+      const signedData = await signer(document, proof);
```
brianorwhatever, line 47: `main` merged the signer's return: `{ ...proofTemplate, ...signedData.proof }`, so a caller's signer could supply or override `verificationMethod`, `created`, `id`. This adapter takes only `proofValue` and drops the rest. `createWitnessProof` is exported and takes a user callback, so third-party witness signers relying on the merge break silently. Intentional?

src/did-document.ts
```diff
@@ -0,0 +1,271 @@
...
+      serviceEndpoint: `${baseUrlWithTrailingSlash}whois.vp`,
+    });
+    changed = true;
+  }
+
+  return changed ? { ...doc, service: services } : doc;
+}
+
+export function generateParallelDidWeb(didwebvhDid: string, didwebvhDoc: DIDDoc): DIDDoc {
+  let webDoc = addDefaultDidWebvhServices(didwebvhDid, deepClone(didwebvhDoc), { idStyle: 'fragment' });
```
brianorwhatever, line 253: Two deltas from `main`'s inline version: the existence check went from `id.endsWith('#files')` to exact `#files` / `${did}#files`, so a doc already declaring `did:web:example.com#files` now gets a duplicate implicit service; and endpoints come from `getBaseUrl()`, which doesn't percent-decode path segments where `main` ran `decodeURIComponent` over the whole domain+path. The second is more consistent with the resolver, but both change emitted `serviceEndpoint` values — a test over a pathed/percent-encoded DID would lock them down.

src/method_versions/method.v1.0.resolution.ts
```diff
@@ -0,0 +1,585 @@
...
+    resolverContext.meta.updateKeys = parameters.updateKeys ?? [];
+  }
+  if (parameters.deactivated === true) {
+    resolverContext.meta.deactivated = true;
+  }
+  if (hasOwn(parameters, METHOD_PARAMETER_KEYS.nextKeyHashes)) {
+    resolverContext.meta.nextKeyHashes = parameters.nextKeyHashes ?? [];
+    resolverContext.meta.prerotation = resolverContext.meta.nextKeyHashes.length > 0;
+  }
+  const normalizedWitness = resolveWitnessParameter(parameters);
```
brianorwhatever, line 482: `resolveWitnessParameter` returns `parameters.witness ?? {}`, so a `witness: null` parameter leaves `meta.witness` as `{}` where `main` assigned `null` directly. No effect on enforcement, but it changes the emitted `didDocumentMetadata.witness`. Intended?

src/utils.ts
```diff
@@ -610,197 +480,114 @@ export async function fetchLogFromIdentifier(identifier: string, controlled: boo
...
+  return {
+    did: parsedDid.did,
+    fragment,
+    keyMultibase: parsedDid.keyMultibase,
   };
+}
 
-  if (!verificationMethod || verificationMethod.length === 0) {
-    return all;
+const findVerificationMethodInDoc = (doc: DIDDoc, vmId: string) => {
```
brianorwhatever, line 565: This is a second copy of `findVerificationMethod`, which now also lives in `did-document.ts:140` — `main` had one. Only this private one is used in `src`; the exported one is reached solely by tests. If the `utils → method → did-document` import cycle is what blocks consolidating, a comment saying so would save the next reader the trip.

vitest.config.ts
```diff
@@ -5,5 +5,14 @@ export default defineConfig({
     environment: 'node',
     include: ['test/**/*.test.ts'],
     exclude: ['node_modules', 'dist'],
+    coverage: {
```
brianorwhatever, line 8: No `coverage.thresholds` here and `test:coverage` isn't wired into CI, so the >90% won't hold on its own.

brianorwhatever (commented):
Nice decomposition — I diffed the branch against `main` as a whole rather than commit-by-commit, and the resolution logic (validation order, meta shaping, snapshot/fallback semantics) comes through faithfully. Comments below are behaviour changes I couldn't find mentioned in the description, mostly questions about intent rather than confirmed bugs.

Also good to see: `enforceRequiredWitnessChecks` no longer mutates the caller's `options.witnessProofs`, and update entries now validate path segments, which `main` skipped entirely.

src/index.ts
brianorwhatever: This doesn’t really matter since we are breaking changes elsewhere and will be creating major. Please ignore

vitest.config.ts
brianorwhatever: thats fine.. ignore this comment

src/method_versions/method.v1.0.resolution.ts
brianorwhatever: We should do what the other libraries are doing here which might be this?

brianorwhatever:
Sorry for the AI comments I was too exhausted yesterday to parse them myself. I’ve gone through them myself now and I don’t think any are blockers. I’d say address what you feel might be helpful and merge when you’re ready

brianorwhatever (approved):

vitest.config.ts
rmlearney-digicatapult: Thanks :) Adding coverage requirements this early could break future work

src/method_versions/method.v1.0.resolution.ts
rmlearney-digicatapult: Thanks Brian, spec says:

* The JSON `null` value **MUST NOT** be used to indicate default or deactivated values, as it removes the typing information required for proper interpretation

* Some early `did:webvh` implementations used the JSON `null` value to indicate the deactivation of parameters such as `watchers`, `witness`, `updateKeys`, `nextKeyHashes`, and `ttl`. Although this usage is deprecated and not valid per the current specification, resolver implementations **SHOULD** gracefully accept `null` and immediately convert the value to their equivalent default value for the parameter when processing [[ref: DID Log entries]]

* For witness specifically, the default/deactivation semantic is `{}`

Moving to `{}` is therefore more spec-aligned. We also perform the necessary conversion from `null` to `{}` in `witness.ts`

src/index.ts
rmlearney-digicatapult: Have renamed the PR, am tracking breaking changes and will update README as well in follow-up

src/did-document.ts
rmlearney-digicatapult: Thanks Brian, in addition to the single-pass improvement you identified, the VM change was meant as a security improvement so that no explicit `authentication` means no silent privilege escalation from incomplete VM input

src/did-document.ts
rmlearney-digicatapult: I will fix this in a follow-up PR

src/witness.ts
rmlearney-digicatapult: The goal was to ensure that `didwebvh-ts` owns the policy and shape of critical proof fields and normalization so that the signer only owns the signing operation. This prevents buggy or malicious signers from mutating `verificationMethod`, `created` or `id` after they have been locked in by the caller of `createWitnessProof`.

* Identity confusion attack
The callback could set `verificationMethod` to a different `did:key` than the one the library intended, so the proof appears to come from a different witness identity.
Impact: attribution errors, incorrect witness counting, and policy bypass attempts.

* Cross-entry replay shaping
The callback could inject an `id` or other proof metadata values that collide with existing proofs, making deduplication or proof-tracking logic ambiguous.
Impact: hard-to-detect replay or overwrite-like behavior in downstream systems that index proofs by id.

* Input-output mismatch exploitation
The library prepares one proof template, but the callback returns a different envelope. That breaks the trust boundary between what was asked to be signed and what was emitted.
Impact: non-deterministic behavior, hard-to-audit signatures, and latent interoperability bugs.

src/utils.ts
rmlearney-digicatapult: I'll try to consolidate in a follow-up PR

swcurran:
w00t!

PR #159, opened 2026-07-28
refactor(core): deprecate and remove v0.5

rmlearney-digicatapult:
Introduce backwards compatibility in `resolve`, forwards only in `create`/`update`/`deactivate`

* `v1.0` resolver now capable of resolving `v0.5` log entries, with special handling of `nextKeyHashes` and `witnesses` for `v0.5` entries but all others consistent with `v1.0`
* Now able to process mixed version logs where later entries differ (greater) than genesis version
* All new `update` and `deactivate` actions work on a `v0.5` log but append a `v1.0` entry
* `Create` no longer allows creation of `v0.5`, only `v1.0`
* All covered by new test suite for backwards compatibility

Other fixes also covered by new regression tests
* Genesis witness `null` normalisation to `{}` for spec compliance
* `#files` and `#whois` fragment de-duplication
* `alsoKnownAs` deduplication

Next steps after this PR:
* file renaming & consolidation of `method.ts` and `method.v1.0.ts`
* separate out a helper file for `verificationMethods` to centralise from across `utils.ts` and `did-document.ts` and eliminate duplicate `findVerificationMethod` caused by circular import path
* remove envars and disk i/o from runtime library, final cleanup of `utils.ts`
* complete migration guide and changelog


swcurran:
Ran the test-suite on this PR and no regression -- the test suite was obviously not using the 0.5 endpoint(s).

Have you created a resolver test that has a 0.5 compliant DID and then becomes a 1.0 DID?   That would be good to have captured -- for a while.  What do you think?

rmlearney-digicatapult:
Hi @swcurran that should be the `Valid mixed-version chain resolves to the v1.0 final state` test inside the new `backwards-compat.test.ts` suite. Generate a `0.5` genesis, append a `0.5` entry, then `1.0` transition and another `1.0` entry

dmitrizagidulin (approved):
👍

dmitrizagidulin (changes requested):
This branch introduces a regression in `updateKeys` carry-forward. In `processV1SubsequentEntry`:                                                                                       
                                                                                                                                                                                          
  ```ts                                                                                                                                                                                   
  if (activeMethod === METHOD_PROTOCOL_V1_0 || hasOwn(parameters, METHOD_PARAMETER_KEYS.updateKeys)) {                                                                                    
    resolverContext.meta.updateKeys = parameters.updateKeys ?? [];                                                                                                                        
  }                                                                                                                                                                                       
  ```                                                                                                                                                                                     
                                                                                                                                                                                          
  The first disjunct is always true for ordinary v1.0 logs, so an update entry that omits `updateKeys` (which `prepareUpdateEntry` produces when the caller passes none and pre-rotation is off) resets the active keys to `[]` instead of carrying them forward.                                                                                                                
                                                                                                                                                                                          
  Repro: `createDID` then two `updateDID` calls with no explicit `updateKeys` -- the second fails with `Key did:key:... is not authorized to update`. Passes on `main`, fails on this branch. Resolution of existing valid logs breaks the same way, and `didDocumentMetadata.updateKeys` is misreported as `[]` even when the omitting entry is last. The existing suite misses it because every multi-update test re-passes `updateKeys` explicitly.                                                                                                            
                                                                                                                                                                                          
Fix: revert to plain `hasOwn(...)` -- the old v0.5 resolver's `if (parameters.updateKeys)` behaves identically for present arrays, so one check serves both versions.

dmitrizagidulin:
The backwards-compat suite doesn't cover logs that 2.x actually produced -- `buildV05Genesis` / `appendLogEntry` synthesize v0.5 logs using v1.0 hash-chaining and timestamp rules. I tried out a real v0.5 log on `main` (`createDID` with `method: 'did:webvh:0.5'` plus one `updateDID`; `main` resolves it fine) and fed it to this branch's resolver. It's rejected:     
                                                                                                                                                                                          
  - `Hash chain broken at '2-...'` -- 2.x's v0.5 `updateDID` hashes update entries over `versionId: '{SCID}'` (see the deleted `method.v0.5.ts`), while the unified resolver recomputes   
  the chain hash over the previous entry's `versionId`.                                                                                                                                   
  - If the entries share a second: `versionTime for version '2' must be greater than previous entry time` -- 2.x had no timestamp ordering check for v0.5.                                
                                                                                                                                                                                          
So resolution backwards compatibility fails for this library's own historical output. Either the unified resolver needs v0.5-specific hash/timestamp handling for pre-transition entries, or the compat claim should be scoped to "spec-conformant 0.5 logs" (noting that 2.x's own output wasn't hash-conformant). Either way, the test fixtures should include at least one log generated by the actual 2.x code path rather than synthesized ones.

Other comments:

- Same-version re-declaration is now a hard error ("redundantly re-declares the already active method"). Previously method: did:webvh:1.0 restated in a later entry was accepted; now it invalidates the log. The spec says method MAY appear to signal a version change, but a hard error on a harmless restatement is an interop risk for logs from lenient writers, so just wanted to check if it's ok.
- `hasServiceFragment` loosened from exact-id match to `endsWith('#files')` - a service whose id belongs to a different DID but ends in `#files` now suppresses the implicit service; also sits oddly next to the migration checklist item "review service endpoint filtering for exact-id matching".

rmlearney-digicatapult:
@dmitrizagidulin you correctly picked up on something that concerned me about the test generator. Do you happen to have any correctly structured 0.5 logs we could add as test fixtures, or advice on fixing the generator to correctly reproduce entries?

rmlearney-digicatapult:
@dmitrizagidulin have addressed:
* carry-forward regression + new tests
* v0.5 fixtures not using real 2.x hash chaining by adhering to the v0.5 spec and not the 2.x implementation quirks
* dropping hard error for same-version re-declaration
* removed false positive `endsWith` for implicit service injection

dmitrizagidulin (approved):
Thanks again!

PR #163, opened 2026-07-30
refactor(core): address circular import of verificationMethod helpers

rmlearney-digicatapult:
Fix circular import by splitting `resolveVM()` into specific `did:key` and `did:webvh` helpers used at call sites

Refactored split of VM helpers across codebase by consolidating others into single `utils/verification-methods.ts` file


dmitrizagidulin (approved):
👍

dmitrizagidulin (dismissed):
this branch inherits the `updateKeys` carry-forward regression reported on #159 (it's in the base commits, not this PR's delta) -- should resolve itself once #159 is fixed and this is rebased.

dmitrizagidulin:
Minor note on the PR description: the import cycle isn't actually eliminated, just relocated. `resolveVM` still calls `resolveDIDFromLog`, so the cycle is now `method.ts -> method_versions/method.v1.0.ts -> assertions.ts -> utils/verification-methods.ts -> method.ts`.
It DOES remove the duplicated `findVerificationMethod` and gets `utils.ts` out of the cycle, which is really helpful.
Completely breaking the cycle would mean injecting the resolver into `resolveVM` (e.g. as a parameter) rather than importing it. Fine to defer that; just flagging so the changelog/description stays accurate.

rmlearney-digicatapult:
@dmitrizagidulin fixed the issue by splitting `resolveVM` into a specific helper for `did:webvh` and moving this into the `method.v1.0.resolver.ts`. Callers now specifically either call the `did:key` parser or the new `did:webvh` helper.

The `resolve-http.test.ts` suite was edited to match, and coverage remains high. Added one regression test for `deepClone()` DateTime handling.

swcurran (approved):
Tested using the test suite and went through an AI review.  LGTM!

brianorwhatever:
Looks good to me. The only things that came up in review were that reaolveDidWebvhVerificationMethod is dead code now and a redundant parse on witness.ts:265 but those can both be included in a cleanup. Merge ready imo

brianorwhatever (approved):

PR #165, opened 2026-08-12
fix(resolver): fix missing didDocumentMetadata fields

rmlearney-digicatapult:
Running `didwebvh-test-suite` identified two non-fatal spec diffs which are addressed in this PR

* missing `versionTime` on resolution
* `ttl` now defaults to `3600` if missing, historic `null` handled
* deactivated DIDs now return `deactivated=true` when resolving historic entries


swcurran (approved):
LGTM and passes the test suite.  Leaving it to you @rmlearney-digicatapult to merge.

PR #167, opened 2026-08-25
types: declare address/services/paths on UpdateDIDInterface

sumitvekariya:
## Problem

`services`, `address` and `paths` are supported on the update path, but they are not declared on `UpdateDIDInterface`. Instead the identical three-field intersection is repeated in four places:

- `src/method.ts` — the exported `updateDID`
- `src/method_versions/method.v1.0.ts` — `updateDID`
- `src/method_versions/method.v1.0.entries.ts` — `prepareUpdateEntry` (twice)

```ts
options: UpdateDIDInterface & {
  services?: ServiceEndpoint[];
  address?: string;
  paths?: string[];
}
```

`CreateDIDInterface` already declares all three directly, so the two entry points differ for no apparent reason.

The practical effect: a caller who types a value as `UpdateDIDInterface` (rather than inlining the object at the call site) cannot set `services` without a cast — even though `prepareUpdateEntry` reads `options.services` and assigns it to `doc.service`.

## Change

Move the three fields onto `UpdateDIDInterface` and delete the four duplicated intersections, which also stops the copies drifting apart. Now-unused `ServiceEndpoint` imports dropped, and `biome check --write` applied to the two files whose signatures reflowed.

## Verification

Types only, no behaviour change. `tsc --noEmit`, `biome check .` and `vitest run` produce **identical** results before and after.

For transparency: this environment has two pre-existing failures unrelated to the change — a `json-canonicalize` module-resolution error under `tsc`, and one base58 assertion — both present on the untouched tree. Test totals are identical either way (20 failed files / 7 passed, 1 failed test / 60 passed).

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,0 +1,369 @@
+{
+  "lockfileVersion": 1,
+  "workspaces": {
+    "": {
+      "name": "didwebvh-ts",
+      "dependencies": {
+        "@noble/curves": "^2.2.0",
+        "@noble/hashes": "^2.2.0",
+        "@scure/base": "^2.2.0",
+        "cookie": "^1.1.1",
+        "did-resolver": "^5.0.1",
+        "glob": "^13.0.6",
+        "js-yaml": "^5.2.1",
+        "json-canonicalize": "^2.0.0",
+      },
+      "devDependencies": {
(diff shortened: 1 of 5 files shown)


rmlearney-digicatapult (approved):
👍

PR #170, opened 2026-08-27
fix(interface): hide private internal requestedDID, expose public witnessProofs

rmlearney-digicatapult:
Security hardening

`requestedDid` is used for internally checking the DID-to-log binding and should not be available to a caller

`witnessProofs` moved onto the interface to clean up API method code


swcurran (approved):
LGTM -- looked at the code (as best I can), AI check/summary review, and the test-suite passes.  All good.

PR #171, opened 2026-08-27
fix(core): CLI - runtime separation

rmlearney-digicatapult:
* Remove paths which propagate envars into runtime
* CLI now owns filesystem logic
* Added `resolveControlledDid` callback to `ResolutionOptions` interface to enable caller/wrapper to pass locally controlled DID Log into runtime. Used by the CLI along with its new envar handling logic.


brianorwhatever (approved):
Clean separation, and the package-boundary test is a nice guard against regressions. LGTM.

PR #172, opened 2026-08-27
feat(cli): CLI error handling

rmlearney-digicatapult:
CLI now wraps and reports runtime errors

CLI tests moved to live inside `cli/` and build script updated


src/cli/index.ts
```diff
@@ -262,49 +289,48 @@ export async function handleResolve(args: string[]) {
...
-
-    console.log('Resolved DID:', did);
-    console.log('DID Document:', JSON.stringify(doc, null, 2));
-    console.log('Metadata:', JSON.stringify(meta, null, 2));
-
-    return { did, doc, meta };
   } catch (error) {
-    console.error('Error resolving DID:', error);
-    process.exit(1);
+    throw new CliError(`Error resolving DID: ${error instanceof Error ? error.message : String(error)}`);
```
brianorwhatever, line 342: The `Resolution error` throws sit inside the `try`, so this `catch` rewraps a `CliError` that was already a `CliError`. Running the branch:

```
$ pnpm cli resolve --log bad.jsonl
Error resolving DID: Resolution error: {
  "error": "invalidDid",
  ...
```

It also drops the inner `exitCode`, so the constructor parameter is effectively dead. One line at the top of each handler's `catch` fixes it:

```ts
if (error instanceof CliError) throw error;
```

Two minor asides while you're in there, non-blocking: `main`'s `default` case no longer prints usage on an unknown command (it did on main), and the top-level `catch` prints only `message`, which hides the stack for non-`CliError` bugs.

brianorwhatever (commented):
One small fix needed; otherwise this is a good change.

src/cli/index.ts
rmlearney-digicatapult: Thanks @brianorwhatever have addressed the re-wrapping issue and the regressions. Have also added a test suite for error paths to ensure these are caught again in future.

src/cli/index.ts
```diff
@@ -691,22 +691,22 @@ export async function main() {
...
-        break;
+        return 0;
       case 'resolve':
         await handleResolve(args);
-        break;
+        return 0;
       case 'verify-proofs':
         await handleVerifyProofs(args);
-        break;
+        return 0;
```
brianorwhatever: `handleVerifyProofs()` can set `process.exitCode = 1` which will be overwritten by `0` here.. not sure the best way to fix this

src/cli/__tests__/error-handling.test.ts
```diff
@@ -0,0 +1,311 @@
...
+        // but capture the log for use with handleDeactivate against a clean environment.
+        const logSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
+        const created = await handleCreate(['--address', 'example.com', '--portable']);
+        logSpy.mockRestore();
+        fs.writeFileSync(logFile, `${created.log.map((entry) => JSON.stringify(entry)).join('\n')}\n`);
+
+        // Ensure both the in-process env var and the on-disk .env fallback are clear,
+        // since getVerificationMethodsFromEnv() reads whichever is present.
+        delete process.env.DID_VERIFICATION_METHODS;
+        fs.writeFileSync(ENV_FILE, '');
```
brianorwhatever: I think this might overwrite the same `.env` as the e2e tests. We will need some isolation here

brianorwhatever (approved):
Apologies for taking so long on this review. Claude found a couple things.. Tests are probably fine as is but the verification  exit code should probably be fixed

src/cli/__tests__/error-handling.test.ts
rmlearney-digicatapult: Thanks @brianorwhatever have isolated `.env` into its own temp directory per test

src/cli/index.ts
rmlearney-digicatapult: Now using `CliError` to return error

rmlearney-digicatapult:
Thanks @brianorwhatever have made changes to test isolation and error propagation

PR #173, opened 2026-09-04
feat(controller): controller APIs for witnessing convenience

rmlearney-digicatapult:
Enabling `Controller` <-> `Witness` agent interactions requires 2 new convenience API methods to be exposed for `controller` operations

* `getWitnessRequirements` enables a `controller` to determine exactly which proofs to obtain for a proposed log chain tip before publishing `did-witness.json` and then `did.jsonl`.

* `verifyWitnessProofs` enables a `controller` to verify returned witness proofs against their did log to ensure validity and threshold without network fetch. All other verification failures (hash chain, SCID, controller proof, etc.) still throw.

Both exported from package root with corresponding types

Other changes:

* `updateDID` and `deactivateDID` now do not fetch witness proofs when `witnessProofs` is omitted, they pass `[]` instead, avoiding network fetch and causing an unmet requirement to fail immediately and deterministically.
* callers must explicitly supply `witnessProofs` when updating/deactivating a DID with an active witness requirement.
* Witness requirement derivation is extracted from the resolver into a shared core module, so resolution and the new public APIs apply the same transition rules.
* `resolveDID` and `resolveDIDFromLog` retain their normal deterministic witness-file fetch behaviour when proofs are omitted.

Tests cover valid, missing, invalid, duplicate, and cumulative witness proofs, transition cases and the deactivation rule

### DIFF
--- a/src/core/resolution.ts
+++ b/src/core/resolution.ts
@@ -32,7 +32,7 @@ import {
   resolveWitnessParameter,
   validateWitnessParameter,
 } from '../witness';
-import { getRequiredWitnessForEntry, type RequiredWitnessCheck } from './witness-requirements';
+import { getRequiredWitnessForEntry, type RequiredWitnessCheck, type WitnessCheckResult } from './witness-requirements';
 
 const hasOwn = <K extends PropertyKey>(obj: object, key: K): obj is Record<K, unknown> => Object.hasOwn(obj, key);
 
@@ -74,6 +74,10 @@ const isSupportedInitialMethod = (m: string | undefined): m is string =>
 
 type InternalResolutionOptions = ResolutionOptions & {
   requestedDid?: string;
+  // Internal-only hook used by `verifyWitnessProofs` (see method.ts) to observe
+  // per-entry witness check outcomes without converting an unmet threshold
+  // into a thrown error.
+  onWitnessChecksComputed?: (checks: WitnessCheckResult[]) => void;
 };
 
 export const resolveLog = async (
@@ -583,6 +587,7 @@ const finalizeResolutionChecks = async ({
       onThresholdFailure: () => {
         resolverContext.witnessThresholdFailure = true;
       },
+      onWitnessChecksComputed: options.onWitnessChecksComputed,
     });
   }
 };
@@ -594,6 +599,7 @@ const enforceRequiredWitnessChecks = async ({
   logEntries,
   verifier,
   onThresholdFailure,
+  onWitnessChecksComputed,
 }: {
   requiredWitnessChecks: RequiredWitnessCheck[];
   witnessProofs: WitnessProofFileEntry[] | undefined;
@@ -601,6 +607,7 @@ const enforceRequiredWitnessChecks = async ({
   logEntries: DIDLog;
   verifier: ResolutionOptions['verifier'];
   onThresholdFailure: () => void;
+  onWitnessChecksComputed?: (checks: WitnessCheckResult[]) => void;
 }): Promise<void> => {
   let resolvedWitnessProofs = witnessProofs;
   if (!resolvedWitnessProofs) {
@@ -608,6 +615,13 @@ const enforceRequiredWitnessChecks = async ({
   }
 
   const publishedVersionNumbers = new Map(logEntries.map((entry, index) => [entry.versionId, index + 1]));
+
+  // When `onWitnessChecksComputed` is supplied (internal-only, see
+  // `verifyWitnessProofs` in method.ts), every check is computed and reported
+  // to the caller instead of throwing on the first unmet threshold, so the
+  // caller can report a complete per-entry breakdown rather than stopping at
+  // the first failure.
+  const computedChecks: WitnessCheckResult[] = [];
 
   for (const check of requiredWitnessChecks) {
     const candidateProofs = resolvedWitnessProofs.filter((witnessProof) => {
@@ -622,12 +636,25 @@ const enforceRequiredWitnessChecks = async ({
       verifier
     );
     const threshold = normalizeWitnessThreshold(check.witness.threshold);
+    const satisfied = approvals >= threshold;
 
-    if (approvals < threshold) {
+    if (onWitnessChecksComputed) {
+      computedChecks.push({ ...check, approvals, satisfied });
+      if (!satisfied) {
+        onThresholdFailure();
+      }
+      continue;
+    }
+
+    if (!satisfied) {
       onThresholdFailure();
       throw new Error(
         `Witness threshold not met for version ${check.targetVersionId}: got ${approvals}, need ${check.witness.threshold}`
       );
     }
+  }
+
+  if (onWitnessChecksComputed) {
+    onWitnessChecksComputed(computedChecks);
   }
 };
--- a/src/core/resolution.ts
+++ b/src/core/resolution.ts
@@ -13,7 +13,6 @@ import type {
   DIDLogEntry,
   DIDResolutionMeta,
   ResolutionOptions,
-  WitnessParameterResolution,
   WitnessProofFileEntry,
 } from '../interfaces';
 import { buildProblemDetails } from '../resolver-result';
@@ -29,19 +28,13 @@ import { MAX_FUTURE_SKEW_MS, parseUtcIso8601VersionTime } from '../utils/iso8601
 import {
   countVerifiedWitnessApprovals,
   fetchWitnessProofs,
-  hasActiveWitnessRequirement,
   normalizeWitnessThreshold,
   resolveWitnessParameter,
   validateWitnessParameter,
 } from '../witness';
+import { getRequiredWitnessForEntry, type RequiredWitnessCheck } from './witness-requirements';
 
 const hasOwn = <K extends PropertyKey>(obj: object, key: K): obj is Record<K, unknown> => Object.hasOwn(obj, key);
-
-interface RequiredWitnessCheck {
-  targetVersionId: string;
-  targetVersionNumber: number;
-  witness: WitnessParameterResolution;
-}
 
 interface ResolutionSnapshot {
   did: string;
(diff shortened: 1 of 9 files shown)


dmitrizagidulin:
With `witnessProofs` now defaulting to `[]`, what happens to `pnpm cli -- update` and `pnpm cli -- deactivate` on a DID that has witnesses? As far as I can tell, `src/cli/index.ts` calls `updateDID` and `deactivateDID` without passing `witnessProofs`, and `--witness-file` is only wired into `resolve`. Before this change those commands fetched the published `did-witness.json`; after it they would fail with "Witness threshold not met". Was that intended, and should `--witness-file` be threaded into `update`/`deactivate` here, or should the fetch fallback stay when the option is omitted?

src/method.ts
```diff
@@ -240,7 +255,10 @@ export const deactivateDID = async (
 ): Promise<{ did: string; doc: DIDDoc; meta: DIDResolutionMeta; log: DIDLog }> => {
   const log = options.log;
   const lastEntry = log[log.length - 1];
-  const lastMeta = (await resolveLog(log, { verifier: options.verifier })).meta;
+  // Default to an empty proof array so missing witnessProofs never falls back to network fetch.
+  // Unmet witness requirement surfaces as a thrown error below.
+  const lastMeta = (await resolveLog(log, { verifier: options.verifier, witnessProofs: options.witnessProofs ?? [] }))
```
dmitrizagidulin: Should this be treated as a breaking change? `deactivateDID` did not accept `witnessProofs` before, so every existing caller with a witnessed DID was relying on the fetch of the published `did-witness.json`, and `updateDID` callers who omitted the option were too. With the `[]` default those calls now fail. Since the CHANGELOG is generated from commit messages, would you add a `BREAKING CHANGE:` footer, or alternatively make this opt-in: keep the fetch when the option is `undefined` and treat an explicit `witnessProofs: []` as "no fetch"? A controller who wants the deterministic, network-free path can already pass `[]` today, so the opt-in route seems to get the same benefit without the compatibility cost.

src/core/resolution.ts
```diff
@@ -647,12 +636,25 @@ const enforceRequiredWitnessChecks = async ({
...
+      if (!satisfied) {
+        onThresholdFailure();
+      }
+      continue;
+    }
+
+    if (!satisfied) {
       onThresholdFailure();
       throw new Error(
         `Witness threshold not met for version ${check.targetVersionId}: got ${approvals}, need ${check.witness.threshold}`
```
dmitrizagidulin: The error might be misleading to devs here. When `updateDID` or `deactivateDID` is called on a witnessed DID without `witnessProofs`, the new `[]` default means this surfaces as "Witness threshold not met for version X: got 0, need N". The actual problem in that case is that the caller omitted the option. If the `[]` default stays, could update/deactivate detect the omitted option up front and throw something like "witnessProofs is required when the DID has an active witness configuration", leaving this message for the genuine short-of-threshold case?

src/core/witness-requirements.ts
```diff
@@ -0,0 +1,99 @@
...
+ * Walks a DID log and derives the witness approvals required for each entry,
+ * applying the did:webvh witness transition rules (genesis activation,
+ * inheritance, replacement, and removal). Performs no cryptographic
+ * verification and requires no `Verifier`; only structural witness-parameter
+ * validation via `validateWitnessParameter`.
+ *
+ * This is the pure, synchronous, verifier-independent counterpart to the
+ * resolver's per-entry witness-requirement tracking. `resolveLog` and
+ * `getWitnessRequirements` both derive their required-witness checks from
+ * this single function.
```
dmitrizagidulin: Small doc nit: as far as I can tell, `resolveLog` does not call `computeWitnessRequirementChecks`. It shares `getRequiredWitnessForEntry`, but keeps its own walk of `meta.witness` inside the genesis/subsequent entry processors, so the genesis-vs-inherit-vs-explicit-`{}` derivation now lives in two places. Would it be worth either having the resolver consume this function too, or softening the comment here (and the one above on `getRequiredWitnessForEntry`) so it describes what is actually shared? Mostly worried about the two walks drifting later.

src/method.ts
```diff
@@ -276,3 +294,67 @@ export const deactivateDID = async (
...
+    onWitnessChecksComputed: (checks) => {
+      checkOutcomes = checks.map((check) => ({
+        targetVersionId: check.targetVersionId,
+        approvals: check.approvals,
+        satisfied: check.satisfied,
+      }));
+    },
+  });
+
+  const outcomesByVersionId = new Map(checkOutcomes.map((outcome) => [outcome.targetVersionId, outcome]));
```
dmitrizagidulin: Could this join be avoided? `WitnessCheckResult` already extends `RequiredWitnessCheck`, so the callback output carries the governing witness config for each entry. If the annotated requirements were built from `checkOutcomes` alone, `verifyWitnessProofs` would not depend on two independent derivations agreeing by `versionId`. As written, a mismatch between the pure walk and the resolver would silently show up as `approvals: 0, satisfied: false` rather than an error.

src/core/resolution.ts
```diff
@@ -81,6 +74,10 @@ const isSupportedInitialMethod = (m: string | undefined): m is string =>
 
 type InternalResolutionOptions = ResolutionOptions & {
   requestedDid?: string;
+  // Internal-only hook used by `verifyWitnessProofs` (see method.ts) to observe
+  // per-entry witness check outcomes without converting an unmet threshold
+  // into a thrown error.
+  onWitnessChecksComputed?: (checks: WitnessCheckResult[]) => void;
```
dmitrizagidulin: Just a thought on shape: the callback both changes control flow (suppresses the throw) and carries data out through a closure, which took me a moment to follow. Would an internal flag like `collectWitnessChecks` with the checks returned on the resolution result, or a small internal helper that both `enforceRequiredWitnessChecks` and `verifyWitnessProofs` call, read more plainly? Not a blocker, happy to leave it if you prefer this.

src/method.ts
```diff
@@ -276,3 +294,67 @@ export const deactivateDID = async (
...
+    const outcome = outcomesByVersionId.get(requirement.versionId);
+    return {
+      ...requirement,
+      approvals: outcome?.approvals ?? 0,
+      satisfied: outcome?.satisfied ?? false,
+    };
+  });
+
+  return {
+    verified: annotatedRequirements.every((requirement) => requirement.satisfied),
```
dmitrizagidulin: Possibly a follow-up rather than this PR: a controller using this API would probably want to know why a proof was discarded (unknown witness, bad signature, duplicate), and right now that only reaches `console.warn` inside `countVerifiedWitnessApprovals`. Would you be open to an issue for surfacing rejected proofs in the result?

src/interfaces.ts
```diff
@@ -270,3 +271,35 @@ export interface WitnessProofFileEntry {
   versionId: string;
   proof: DataIntegrityProof[];
 }
+
+/**
+ * Common structural input accepted by the witness verification APIs.
+ * Compatible with any method exposing `log: DIDLog`.
+ */
+export interface WitnessVerifiableResult {
```
dmitrizagidulin: Minor API question: `resolveDIDFromLog` takes a bare `DIDLog`, so would these two functions be more consistent taking `DIDLog` as well (callers pass `result.log`)? Also, the README mentions passing a `DIDResolutionResult`-shaped value, but that type does not have a `log` field, so that example may not type-check.

test/witness.test.ts
```diff
@@ -444,6 +467,432 @@ describe('Witness Implementation Tests', async () => {
...
+
+      requirements[0].witnesses.push({ id: 'did:key:zTamperedWitness' });
+      requirements[0].threshold = 999;
+
+      const requirementsAgain = getWitnessRequirements(initialDID);
+      expect(requirementsAgain[0].witnesses).toEqual(originalWitnesses);
+      expect(requirementsAgain[0].threshold).toBe(2);
+    });
+
+    test('Agrees with the resolver-derived requirements for the same log', async () => {
```
dmitrizagidulin, line 743: I may be misreading this, but does this test end up comparing `getWitnessRequirements` against the function it is a direct `map` over? If so it would pass regardless of what the resolver does. Would it be more useful to compare against the checks the resolver itself produces (for example via the `onWitnessChecksComputed` hook), so the test actually catches the two derivations drifting?

test/witness.test.ts
```diff
@@ -1,12 +1,14 @@
-import { beforeAll, describe, expect, test } from 'vitest';
+import { beforeAll, describe, expect, test, vi } from 'vitest';
+import { getWitnessRequirements, verifyWitnessProofs } from '../src';
+import { computeWitnessRequirementChecks } from '../src/core/witness-requirements';
```
dmitrizagidulin: Tiny observation, not a request: this is the only consumer of `computeWitnessRequirementChecks` outside the `getWitnessRequirements` wrapper, which reaches into `src/core` from the tests. Fine for an internal test, just noting it in case the comparison test above ends up using the resolver output instead, in which case this import could go away.

src/method.ts
rmlearney-digicatapult: Going to review this (and all your comments). One thought I had - should we pass a logger through the codebase so we're not just using console for everything? Probably a post-3.0.0 change

test/witness.test.ts
rmlearney-digicatapult: Have strenghened the resolver-comparison test to consume the resolver generated checks allowing removal of direct core import

test/witness.test.ts
rmlearney-digicatapult: Yes good catch, have now strenghened the resolver comparison test to consume the resolver generated checks

src/method.ts
rmlearney-digicatapult: Done and now covered in tests

rmlearney-digicatapult: Done

src/core/resolution.ts
rmlearney-digicatapult: replaced callback with a split between evaluation (`evaluateRequiredWitnessChecks`) and enforcement (`resolveLog` which consumes these results and throws if requirements unsatisfied) involving a bit more refactoring but resulting in clearer code. Old `resolveLog` split to create `resolveLogWithWitnessResults` which returns evaluated results without throwing.

src/method.ts
rmlearney-digicatapult: I've restored the previous behaviour and made it clear in the README that passing no WitnessProofs (`undefined`) will result in a network fetch, but explicitly passing `[]` will not cause a network fetch but fail-fast if witness proofs are actually required. The `deactivateDID` method not requiring `witnessProofs` was spec-incorrect before so this change results in spec conformity.

src/core/resolution.ts
rmlearney-digicatapult: Have restored previous behaviour accepting `undefined` for network fetch and `[]` for no fetch which also performs fail-fast requirement assessment

src/core/witness-requirements.ts
rmlearney-digicatapult: I plan to refactor to remove the duplication in a follow-up PR. I've added regression coverage to the test suite prior to refactoring.

rmlearney-digicatapult:
@dmitrizagidulin thanks for the review. Have addressed all but one of your comments, saved for a subsequent refactoring PR. Have extended the CLI to enable witness proofs on `update` and `deactivate` and add the controller lifecycle `verify-proofs` functionality.

dmitrizagidulin (approved):

Issue #160, opened 2026-07-29
Finicky pre-export declaration syntax in types declaration


bumblefudge:
Should [this export declaration](https://github.com/decentralized-identity/didwebvh-ts/blob/052ede92057fb444987cdb8abb35ea7e39c115ee/src/types.ts#L6) include the filetype suffix (`.ts`) ?  Claude (Sonnet 5, nothing Mythological) is being a little extra about it:

<img width="807" height="239" alt="Image" src="https://github.com/user-attachments/assets/06606775-27d7-46be-9bf5-09f9c627777e" />

Obviously don't think twice about closing the issue if it's a non-issue or there are valid reasons to omit the suffix, just giving a heads up in case it's both real and a nobrainer to patch.

PR #177, opened 2026-09-07
fix(declarations): address typescript declaration issue

rmlearney-digicatapult:
Address #160 by adding extension checking to `Biome` and adding `.js` extensions to all imports

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -45,6 +45,9 @@
       "includes": ["examples/**", "scripts/**"],
       "linter": {
         "rules": {
+          "correctness": {
+            "useImportExtensions": "off"
+          },
           "complexity": {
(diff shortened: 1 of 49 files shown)


brianorwhatever:
One small fix: the imports in examples/express-resolver.ts and examples/signer.ts should stay as didwebvh-ts and didwebvh-ts/types. Adding .js to those package paths causes ERR_PACKAGE_PATH_NOT_EXPORTED, since they don’t match the exports in package.json.

The .js extensions on relative imports look good. Could you revert just those package imports?

rmlearney-digicatapult:
@brianorwhatever have reverted the .js extensions for express-resolver and updated biome settings

brianorwhatever (approved):

PR #181, opened 2026-09-10
fix: normalize update keys at DID write boundaries

### LINKED ISSUE #166: isKeyAuthorized normalizes the proof's verificationMethod but not updateKeys, so getVerificationMethodId() output can never authorize
`isKeyAuthorized` compares two different representations of the same key and normalizes only one side. The result is that the value returned by this library's own `getVerificationMethodId()` — the natural thing to put in `updateKeys` — can never authorize an entry, and because `createDID`/`updateDID` self-verify before returning, DID creation fails closed with an error that points away from the cause.

Reproduced on **2.8.0** (current release). The code is unchanged on `main` as of this writing.

## The asymmetry

`src/assertions.ts:9`

```ts
const isKeyAuthorized = (verificationMethod: string, updateKeys: string[]): boolean => {
  const parsedVerificationMethod = parseDidKeyVerificationMethod(verificationMethod);

  return updateKeys.some((updateKey) => {
    return updateKey === parsedVerificationMethod.keyMultibase;
  });
};
```

The proof's `verificationMethod` is parsed down to a bare `keyMultibase`. Each `updateKeys` entry is compared **raw**. One key, two encodings, reconciled on only one side of the `===`.

Meanwhile `src/cryptography.ts:123`:

```ts
getVerificationMethodId(): string {
  return this.useStaticId
    ? `did:key:${this.verificationMethod.publicKeyMultibase}#${this.verificationMethod.publicKeyMultibase}`
    : this.verificationMethod.id || '';
}
```

That absolute form is what `createDID`/`updateDID` stamp into `proof.verificationMethod`, and it is also the obvious value to hand to `updateKeys`. Do that and the comparison can never match.

To be clear: requiring bare multikey in `updateKeys` is spec-correct — did:webvh says those are multikeys. The defect is that the library ships an accessor whose output its own verifier rejects, with nothing bridging the two.

## Reproduction

One key, three encodings of it:

```
getVerificationMethodId() -> did:key:z6MkjXzxx…Vctx#z6MkjXzxx…Vctx
bare multibase           -> z6MkjXzxx…Vctx

PASS  bare multibase                 (what the verifier wants)
FAIL  getVerificationMethodId()      (the library's own accessor)
        Key did:key:z6MkjXzxx…Vctx#z6MkjXzxx…Vctx is not authorized to update.
FAIL  did:key: prefix, no fragment
        Key did:key:z6MkjXzxx…Vctx#z6MkjXzxx…Vctx is not authorized to update.
```

<details>
<summary>Self-contained script (didwebvh-ts 2.8.0, @noble/ed25519, @noble/hashes)</summary>

```js
import { createDID, AbstractCrypto, multibaseEncode, MultibaseEncoding, prepareDataForSigning } from "didwebvh-ts";
import * as ed from "@noble/ed25519";
import { sha512 } from "@noble/hashes/sha2.js";
ed.hashes.sha512 = sha512;

const priv = ed.utils.randomSecretKey();
const pub = await ed.getPublicKeyAsync(priv);
const publicKeyMultibase = multibaseEncode(new Uint8Array([0xed, 0x01, ...pub]), MultibaseEncoding.BASE58_BTC);

class Crypto extends AbstractCrypto {
  async sign(input) {
    const data = await prepareDataForSigning(input.document, input.proof);
    return { proofValue: multibaseEncode(await ed.signAsync(data, priv), MultibaseEncoding.BASE58_BTC) };
  }
  async verify(sig, msg, pk) { return ed.verifyAsync(sig, msg, pk); }
}
const mk = () => new Crypto({ verificationMethod: { id: "", type: "Multikey", controller: "", publicKeyMultibase } });
const signer = mk();
const verificationMethods = [{ type: "Multikey", publicKeyMultibase }];

for (const [label, updateKeys] of [
  ["bare multibase", [publicKeyMultibase]],
  ["getVerificationMethodId()", [signer.getVerificationMethodId()]],
  ["did:key: prefix, no fragment", [`did:key:${publicKeyMultibase}`]],
]) {
  try {
    await createDID({ domain: "example.com", signer, verifier: mk(), updateKeys, verificationMethods });
    console.log(`PASS  ${label}`);
  } catch (e) {
    console.log(`FAIL  ${label}\n        ${e.message}`);
  }
}
```

</details>

## Why this is worse than a validation error

Two things compound it.

**It fails closed at creation, not at resolution.** `createDID` and `updateDID` verify the entry they have just built before returning, so the caller gets no log at all rather than a subtly bad one. That is the right design and it is why this is a hard stop rather than silent corruption — but it means a first-time integrator's very first `createDID` throws.

**The error names the one thing that is not wrong.** `Key did:key:z6Mk… is not authorized to update.` identifies the proof's key, which is correctly derived from the private key the caller holds. It mentions neither `updateKeys` nor the fact that the comparison is between two different encodings. It reads as "you are signing with the wrong key", which sends you looking for a lost or rotated controller key. We spent real time on exactly that theory before ruling it out with a fresh keypair and empty storage; the failure is identical for a brand-new identity, which is what finally pointed at the encoding.

## Suggested fix

Normalize both sides — run each `updateKeys` entry through the same parser, tolerating bare multibase, so all three forms above authorize:

```ts
const toKeyMultibase = (key: string): string =>
  key.startsWith('did:key:') ? parseDidKeyVerificationMethod(key).keyMultibase : key;

const isKeyAuthorized = (verificationMethod: string, updateKeys: string[]): boolean => {
  const { keyMultibase } = parseDidKeyVerificationMethod(verificationMethod);
  return updateKeys.some((updateKey) => toKeyMultibase(updateKey) === keyMultibase);
};
```

If you would rather keep `updateKeys` strictly bare — a defensible call, since accepting several encodings for a security-relevant identifier has its own costs — then the error should name the actual mismatch, e.g. report the parsed multibase alongside the `updateKeys` it was compared against, and `getVerificationMethodId()`'s docs should say plainly that its output is not an `updateKeys` value. Either way the current pairing is a trap.

Happy to open a PR for whichever direction you prefer.

brianorwhatever:
Fixes #166.

Passing `signer.getVerificationMethodId()` in `updateKeys` previously failed genesis authorization, and DID-form rotation keys could fail pre-rotation commitments or leave the next update unable to authorize. Normalize supplied keys at `createDID`, `updateDID`, and `deactivateDID` ingress, before entry preparation, hashing, and signing, as recommended in the issue discussion.

- Accept bare multikeys and `did:key` identifiers with an optional matching fragment. Persist bare multikeys in entries and metadata without mutating caller options, key arrays, or existing logs.
- Validate every supplied key using the existing DID-key parsers and the proof-side Ed25519 `0xed01` header check. Reject malformed identifiers and unsupported codecs before signing, identifying the offending `updateKeys` index.
- Keep read-side authorization strict and document the accessor's use with write APIs. Pre-rotation hashes still derive from bare `publicKeyMultibase` values.

Validation: reproduced the original authorization and pre-rotation failures before the fix; added 14 regression cases covering all three forms, all write APIs, invalid unused keys, input preservation, and strict read authorization. Updated the witnessed rotation test that expected the old failure. All 403 tests across 29 files pass, along with `pnpm exec biome ci .`, `pnpm check`, and `pnpm build` (local Node 25.2.1).


test/update-keys.test.ts
```diff
@@ -0,0 +1,139 @@
...
+      verifier: defaultVerifier,
+      signer: createTestSigner(key),
+      verificationMethods: asPublicVerificationMethods(key),
+      updateKeys: [key.publicKeyMultibase!],
+    });
+    for (const form of ['did:key', 'verificationMethod'] as const) {
+      await expect(
+        documentStateIsValid(log[0], [encodeKey(key.publicKeyMultibase!, form)], undefined, true, defaultVerifier)
+      ).rejects.toThrow('is not authorized to update');
+    }
```
rmlearney-digicatapult, line 137: Would be good to follow this with the positive case multikey from forms and assert passing

rmlearney-digicatapult (approved):
Good use of existing helper methods. One nit you could address or ignore.

PR #189, opened 2026-09-13
fix(resolver): guard against localhost HTTP downgrades

### LINKED ISSUE #185: getBaseUrl downgrades to http for any identifier containing "localhost", not just a localhost host
### Summary

`getBaseUrl` selects plaintext `http` when the identifier *contains* the
substring `localhost`, rather than when its host *is* `localhost`. A public
hostname such as `localhost.example.com`, `my-localhost.example.com` or
`cdn-localhost.example` therefore has its DID log fetched over cleartext
HTTP, and so does any identifier with `localhost` anywhere in its path.

dist/esm/index.js (2.7.4), `getBaseUrl`:

```js
let remainder = decodeURIComponent(parts.slice(3).join("/"));
const protocol = remainder.includes("localhost") ? "http" : "https";
```

`remainder` is the host *and* the path, and the test is a substring test, so:

| identifier | fetched |
|---|---|
| `did:webvh:<scid>:localhost%3A8000` | `http://localhost:8000/.well-known/did.jsonl` (intended) |
| `did:webvh:<scid>:localhost.example.com` | `http://localhost.example.com/.well-known/did.jsonl` |
| `did:webvh:<scid>:example.com:localhost` | `http://example.com/localhost/did.jsonl` |

### Impact

The DID log is the source of the update keys, so fetching it over cleartext
lets a network attacker rewrite the document for any of these identifiers —
the hash chain is self-consistent, so a substituted log with its own SCID
fails, but an attacker on the path can still strip or stall the response, and
plaintext defeats the point of serving the log over TLS. It also silently
downgrades a resolver that only ever intended to allow HTTPS: a caller cannot
tell from the identifier that it will be fetched in the clear.

The path case is the surprising half — `example.com:localhost` is a perfectly
ordinary webvh path segment.

### Suggested fix

Decide the scheme from the host only, and only on an exact match:

```js
const [hostPart, ...pathParts] = remainder.split("/");
const [host] = decodeURIComponent(hostPart).split(":");
const isLocal = host === "localhost" || host === "127.0.0.1" || host === "::1";
const protocol = isLocal ? "http" : "https";
```

Ideally the plaintext downgrade is also opt-in (a resolution option, or an
env/config flag), so a library embedding didwebvh-ts can refuse it outright
in production. That pairs with #158: a caller who can inject `fetch` can
enforce its own transport policy, but only if the URL it is handed is not
already downgraded.

### Workaround

We compute the log URL ourselves and call the exported `resolveDIDFromLog`,
treating any did:webvh host that contains `localhost` as local-only unless
the caller explicitly allows private hosts. Happy to send a PR for the
`getBaseUrl` change if the exact-match-plus-opt-in shape looks right.

brianorwhatever:
The `localhost` substring downgrade reported in 2.7.4 is already removed on `main` by 8ebada3d5f90eda24f14c337ef2d3a714f9602ff, which requires HTTPS even for localhost. This PR protects that behavior with regression coverage and corrects the README's obsolete HTTP exception.

Coverage includes the reported public hostnames and path segments, encoded hosts and ports, and complete resolution with assertions on the URL passed to `fetch`. Local testing without HTTPS is documented using `resolveDIDFromLog`.

Fixes #185.

Validation: all 457 tests pass, along with Biome CI, TypeScript checking, and the build. Temporarily restoring the substring-based downgrade makes all 13 added cases fail; after restoring the implementation, all 43 tests in the affected files pass.

### DIFF
--- a/test/resolve-http.test.ts
+++ b/test/resolve-http.test.ts
@@ -76,6 +76,28 @@ describe('resolveDID over HTTPS', () => {
     expect(result.didResolutionMetadata.contentType).toBe('application/did+ld+json');
   });
 
+  test.each([
+    ['localhost.example.com', [], 'https://localhost.example.com/.well-known/did.jsonl'],
+    ['example.com', ['localhost'], 'https://example.com/localhost/did.jsonl'],
+    ['localhost:8000', [], 'https://localhost:8000/.well-known/did.jsonl'],
+  ])('resolves %s with paths %j over HTTPS', async (address, paths, expectedUrl) => {
+    const created = await createDID({
+      address,
+      paths,
+      signer: createTestSigner(authKey),
+      updateKeys: [authKey.publicKeyMultibase!],
+      verificationMethods: asPublicVerificationMethods(authKey),
+      verifier,
+    });
+    const fetchMock = stubFetchResponse(toJsonl(created.log));
+
+    const result = await resolveDID(created.did, { verifier });
+
+    expect(fetchMock).toHaveBeenCalledExactlyOnceWith(expectedUrl);
+    expect(result.didDocument?.id).toBe(created.did);
+    expect(result.didResolutionMetadata.error).toBeUndefined();
+  });
+
   test('ignores DID_VERIFICATION_METHODS in the runtime environment', async () => {
     const previous = process.env.DID_VERIFICATION_METHODS;
     process.env.DID_VERIFICATION_METHODS = 'invalid-runtime-value';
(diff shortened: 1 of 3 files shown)


test/resolve-http.test.ts
```diff
@@ -76,6 +76,28 @@ describe('resolveDID over HTTPS', () => {
     expect(result.didResolutionMetadata.contentType).toBe('application/did+ld+json');
   });
 
+  test.each([
```
rmlearney-digicatapult, line 79: Not as powerful as the second regression test but not a concern

Are these covered in the existing DID transformation tests?

brianorwhatever:
confirms what you mentioned on #185 @rmlearney-digicatapult

rmlearney-digicatapult (approved):

test/resolve-http.test.ts
brianorwhatever: There’s partial overlap: the existing [URL derivation tests](https://github.com/decentralized-identity/didwebvh-ts/blob/a3a6a6c498f8919fbff6f1e64ce0ab8b75d0737f/test/resolve.test.ts) cover HTTPS for localhost, and the [address transformation tests](https://github.com/decentralized-identity/didwebvh-ts/blob/a3a6a6c498f8919fbff6f1e64ce0ab8b75d0737f/test/domain-encoding.test.ts) cover localhost encoding. Neither covers localhost.example.com or example.com:localhost.

These new resolver tests additionally exercise resolution and assert the URL passed to fetch. The other regression table covers URL derivation more broadly.

Issue #180, opened 2026-09-08
A verification method can only hold one purpose, and createVMID is not exported


hacdias:
`VerificationMethod.purpose` is a single `DataIntegrityProofPurpose`, and `normalizeVMs` pushes each method into exactly one relationship. There is no way to express one key that is both `authentication` and `assertionMethod` — the common shape for a DID with a single key.

The workaround is passing explicit `authentication: [...]` / `assertionMethod: [...]` arrays, but those need the verification method's id. `createVMID` derives it as `${did}#${publicKeyMultibase.slice(-8)}`, and at create time the caller doesn't know the DID yet (the SCID isn't derived). `createVMID` also isn't re-exported from `src/index.ts`, so callers either re-implement the fragment derivation or pass a relative `#<fragment>` reference and hope the derivation doesn't change.

credo-ts does the latter: openwallet-foundation/credo-ts#2928

Suggestion: accept `purpose` as an array (or add `purposes`), so one method can land in several relationships without the caller needing to construct its id.

rmlearney-digicatapult:
Changed the way this library handles did documents for 0.3.0. Caller now responsible for passing in a full correctly formatted did document so that we can remove purpose entirely and rely on did document relationships. 

No changes to validation or injection of missing services etc.

Issue #179, opened 2026-09-08
updateDID discards all verification relationships when verificationMethods is supplied


hacdias:
In `prepareUpdateEntry` (`src/core/entries.ts`), supplying `verificationMethods` overwrites all six derived properties from a fresh `createDIDDoc`:

```ts
if (safeVerificationMethods !== undefined) {
  doc.verificationMethod = normalizedUpdateDoc.verificationMethod;
  doc.authentication = normalizedUpdateDoc.authentication;
  doc.assertionMethod = normalizedUpdateDoc.assertionMethod;
  // ... keyAgreement, capabilityDelegation, capabilityInvocation
}
```

`normalizeVMs` only populates a relationship when `vm.purpose` is set, so an update that passes the current `verificationMethod` array straight through — a natural way to change only `service` — silently drops every relationship the DID had, unless the caller also re-passes `authentication` / `assertionMethod` / `keyAgreement`.

On `main` this is broader than in 2.8.0: `normalizeVMs` used to default purpose-less methods into `authentication`, so that one survived. Now none do, and the DID resolves to a document whose keys are authorized for nothing.

We hit this in credo-ts and work around it by re-reading the relationships from the last log state: openwallet-foundation/credo-ts#2928

Suggestion: merge rather than replace — only reset the relationships the caller actually supplies.

rmlearney-digicatapult:
Did document construction now the responsibility of the caller, library now only validates and injects missing services etc. meeting spec requirements.

PR #190, opened 2026-09-13
feat(didDocument): use types from did-resolver

rmlearney-digicatapult:
First part of addressing the issues raised in #179 and #180 

Use the stable `didDocument` types (including `VerificationMethod`, `Service`, etc.) from `did-resolver` instead of our own (faulty) implementations

Our existing methods lack the compulsory `controller` field on `VerificationMethod` and we expose `secretKeyMultibase` which is not part of the DID Core spec

We also use `purpose` when the purpose should be derived from the relationship field in the did document itself but this has been retained in this current PR, for removal in a subsequent PR.

Taking `secretKeyMultibase` out of the interface led to other changes to public-private key separation inside `signer.ts`:

* `generateEd25519VerificationMethod()` renamed to `generateEd25519KeyMaterial()` which now returns
  * `signingKey` (public + private material required by signer)
  * `verificationMethod` (only the public fields allowed in the DID Document

`ExampleCrypto` consumes `signingKey` whereas `createDID()` only receives the public `verificationMethod`

All other signing is unchanged, only the public-private key boundary has been made explicit with a validator update to ensure no signing material enters the DID Document


brianorwhatever (approved):
Looks great! No changes from me. Merge when ready

PR #191, opened 2026-09-14
fix(didDocument): use new did document handling

rmlearney-digicatapult:
One step closer to externalising `didDocument` construction (to match behaviour in the `python` and `rust` implementations)

Multiple interfaces and methods marked with `@deprecated` so that the existing test surface did not require extensive changes - cleaned up in PR #192 

New tests only for did document handling contract


src/cli/index.ts
```diff
@@ -459,33 +476,40 @@ export async function handleUpdate(args: string[]) {
...
 
-    // If we're adding VMs, create a VM for each type
     if (addVm && addVm.length > 0) {
       const vmId = `${did}#${vmPublicKeyMultibase.slice(-8)}`;
-
-      // Add a verification method for each type
-      for (const vmType of addVm) {
-        const newVM: VerificationMethod & { purpose: VerificationMethodType } = {
+      const existingVms = Array.isArray(nextDoc.verificationMethod) ? [...nextDoc.verificationMethod] : [];
+      if (!existingVms.some((existing) => existing.id === vmId)) {
```
brianorwhatever, line 488: Could we normalize relative IDs before comparing here? A caller-authored document could already contain #<key-suffix> (even if this might not be valid anymore, I'm not sure?), while the CLI constructs <did>#<key-suffix>. Adding a relationship then creates a second entry for the same verification method. Since we’re supporting relative IDs in this PR, we should reuse the existing method. The relationship deduplication below has the same issue with includes(vmId)

brianorwhatever (approved):

src/cli/index.ts
rmlearney-digicatapult: Good catch. Have created `did-document.ts` in `cli/` which represents the did document authoring capabilities of a fully-functional caller/wrapper + tests

PR #192, opened 2026-09-14
feat(didDocument): did document managed by controller external to library

rmlearney-digicatapult:
Did Document construction now entirely external to `didwebvh-ts` and required by interfaces. Matches behaviour of `python` and `rust` implementations whereby controllers have full control over `verificationMethods`, key fragments, services.

`webvh` spec requirements unchanged - `alsoKnownAs`, compulsory `service` injection, etc.

Removal of `@deprecated` methods and interfaces. 

`*DIDInterface` renamed `*DIDOptions`.

Tests now use new helper to construct did document rather than passing `verificationMethod` and `purpose`


docs/BREAKING_CHANGES_v3.0.md
```diff
@@ -45,12 +45,12 @@ This document summarizes breaking changes in the 3.0.0 release. For detailed upg
...
 ### Behavior Changes (Semantics/Defaults)
 
-#### 3. Verification Method `purpose` Field
+#### 3. Standard DID Document Authoring (`didDocument` required on `createDID`)
 
-- **Changed**: VMs without `purpose` no longer implicitly enter `authentication`
-- **Now**: `purpose` must be explicitly set
-- **Reason**: Removes ambiguity; makes intent explicit
-- **Upgrade**: Add `purpose: 'authentication'` to auth keys → [Full guide](./UPGRADE_2.x_to_3.0.md#2-verification-method-purpose-field)
+- **Changed**: `createDID` now requires a full, standard W3C `didDocument: DIDDocument` with `{DID}` or `{SCID}` placeholders.
```
swcurran: I wouldn't have expected "{DID}" here -- just "{SCID}".  Presumably the domain portion may be passed in some other way?  OK, just suprising.

swcurran (commented):
LGTM but defer to other Maintainers for their assessment.

My Claude Code review uncovered only this gem:

> [docs/UPGRADE_2.x_to_3.0.md:175](https://github.com/decentralized-identity/didwebvh-ts/blob/3757ab30c194432d5784de6ae737bb6c60aa587f/docs/UPGRADE_2.x_to_3.0.md#L175) — a leftover });/}); fragment from the old example was left behind between migration step 4 and the "Common Patterns" heading. Cosmetic (renders as garbled stray lines in the doc) but worth a quick fix before merge.

That's about as trivial as you can get.

docs/BREAKING_CHANGES_v3.0.md
rmlearney-digicatapult: Rust uses {DID} as a user-convenience method so thought to copy that here. Have expanded the explanation in the doc.

{DID} is expanded from method + address prior to SCID calculation just to make things easier for the caller.

brianorwhatever (approved):
Looks good!

